Issue #3, opened 2016-10-17
Once does not work


mmoutenot:
It is being bound like any other event instead of a special `once` case.

PR #4, opened 2016-10-17
fix once (#2) and queries (#3)

mmoutenot:
`once` cannot be passed to `on`, it is a function unto itself.

the `populates` filter was truthy even when the params were queries because `indexOf` returns an index or `-1` if not present.


prescottprue:
Other than the lint error this looks good. Great catch on the once part, not sure why that got removed.

PR #11, opened 2016-11-22
Updated param name in createAuthProvider()

filimongeorge:
The providerName should be used in the creation of the provider instead of the provider itself


prescottprue:
Great catch! Not sure why it wasn't caught by unit tests. Will release an update later today with the changes included.

PR #17, opened 2016-12-08
Requesting and Requested to state

RahavLussato:
hi,
i did bunch of changes and fixes:

1. remove the isListenToLast - the whole id of queryId is that you want identifier for the query that can be with dynamic path and you want to listen only to the new path, so instead having boolean we can depend only on if we have queryId or not (and also by the way the isListenToLast wasn't expose outside so its fixes that too).

2. add new identifiers : i added new identifiers such isRequesting, isRequested, timestamp that now you can for each query change to have more data, its super useful for example to use timestamp on ShouldComponentUpdate or to have loader while isRequesting true.

3. queryId was broken, now its works again :)

4. i fixed the handling when data parameter was changed for dynamic path, it was relaying of the original path instead of the prev one, so on situations that the path was change it was always comparing to the original one instead of the prev one.


RahavLussato:
@prescottprue looks like your test is not getting path variable so the test is failed

prescottprue:
Starting to check it out now. Did you remove dist from the gitignore file? Could you add it back and remove the git tracking of the dist folder?

RahavLussato:
yes forgot to remove that back.
i added dist, because when i work on my fork i can't npm install that and have npm publish that will build for me.

RahavLussato:
@prescottprue i removed dist :)

src/reducer.js
```diff
@@ -38,24 +40,72 @@ const pathToArr = path => path.split(/\//).filter(p => !!p)
...
-      const { data } = action
+      const { data, rootPath } = action
       pathArr = pathToArr(path)
+      rootPathArr = pathToArr(rootPath)
       retVal = (data !== undefined)
         ? state.setIn(['data', ...pathArr], fromJS(data))
         : state.deleteIn(['data', ...pathArr])
 
+      retVal = (timestamp !== undefined)
+        ? retVal.setIn(['timestamp', ...rootPathArr], fromJS(timestamp))
```
prescottprue: Is there a time where timestamp will ever not be `Date.now()`? If so maybe we should just have this line be `retVal.setIn(['timestamp', ...rootPathArr], fromJS(Date.now()))` and we wouldn't have to pass timestamp into the actions.

src/utils/query.js
```diff
@@ -9,7 +13,7 @@ export const getWatchPath = (event, path) =>
  * @param {String} queryId - Id of query
  */
 export const setWatcher = (firebase, event, path, queryId = undefined) => {
-  const id = queryId ? `${event}:/${queryId}` : getWatchPath(event, path)
+    let id = queryId || getQueryIdFromPath(path) || getWatchPath(event, path)
```
prescottprue: Nit Pick: Is this intentionally a `let` or did you mean to make it a `const`?

src/utils/query.js
```diff
@@ -28,7 +32,7 @@ export const setWatcher = (firebase, event, path, queryId = undefined) => {
  * @param {String} queryId - Id of query
  */
 export const getWatcherCount = (firebase, event, path, queryId = undefined) => {
-  const id = queryId ? `${event}:/${queryId}` : getWatchPath(event, path)
+  let id = queryId || getQueryIdFromPath(path) || getWatchPath(event, path)
```
prescottprue: Nit Pick: Is this intentionally a let or did you mean to make it a const?

src/utils/query.js
```diff
@@ -39,18 +43,18 @@ export const getWatcherCount = (firebase, event, path, queryId = undefined) => {
  * @param {String} path - Path to watch with watcher
  * @param {String} queryId - Id of query
  */
-export const unsetWatcher = (firebase, event, path, queryId = undefined) => {
-  let id = queryId || getQueryIdFromPath(path)
+export const unsetWatcher = (firebase, dispatch, event, path, queryId = undefined) => {
+  let id = queryId || getQueryIdFromPath(path) || getWatchPath(event, path)
```
prescottprue: Nit Pick: Since you no longer reassign id, it can be switched to a `const`.

src/reducer.js
```diff
@@ -70,7 +120,10 @@ export default (state = initialState, action = {}) => {
         authError: null,
         profile: null,
         isLoading: false,
-        data: {}
+        data: {},
+        timestamp: {},
```
prescottprue: Should these parameters also be set in `initialState` const above?

prescottprue:
Great changes! I love the idea of including requesting and requested to state. Doesn't have to be on this PR, but in the future we might want to add helpers or modify `isLoaded` and `isEmpty`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-redux-firebase",
-  "version": "1.1.2",
+  "version": "1.1.6",
```
RahavLussato: sure i just changed that to identify the changes on my fork.

src/reducer.js
RahavLussato: if you can't think about any case that will be delay between the dispatch and the time it will get to the reducer to this line so its ok, but if some delay will occur like on the prev line of setting the data on state (big data maybe ?) the times will not be the same.
its up to you :) for me i always prefer that the reducer will not have control on values like dates and it will come from outside.

src/reducer.js
RahavLussato: yes sure, i missed that. can you add also the others ? (isRequesting: {}, isRequested: {}).

src/utils/query.js
RahavLussato: i'll change it to const

src/utils/query.js
RahavLussato: i'll change it to const

src/utils/query.js
RahavLussato: i'll change it to const :)

RahavLussato:
@prescottprue i did the changes.
i didn't change the timestamp, tell me if you want me to change that.
i also didn't change the initial state, i think its better to stay there.

about the isRequesting and isRequested they are in but not as helpers.
you can fetch them from the state on specific path like :
`@connect(
    (state) => ({
        firebaseTimestamp: customToJS(state.firebase, 'some/path', 'timestamp'),
        firebaseRequesting: customToJS(state.firebase, 'some/path', 'requesting'),
        firebaseRequested: customToJS(state.firebase, 'some/path', 'requested')}),
    {})`

PR #23, opened 2017-01-05
Compatibility with react-boilerplate

michelalbers:
This pull request achieves two things:

1. This ensures compatibility with some boilerplates (e.g. https://github.com/mxstbr/react-boilerplate) by using ```require('firebase/app');``` instead of ```import firebase from 'firebase';```.
2. Lists in Firebase are build as objects with uids as keys likes this:
   #### Firebase response 
   ```JavaScript
    {
        'absdfTGTdd-kV': true,
        'absdfTGTdd-kX': true
    }
    ```
    With this PR the lib checks for value true and takes the key as id value. Now populate works with the above scheme, too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ branches:
   - master
 script:
 - npm run lint:fix
-- npm run test:cov
+# - npm run test:cov
 cache:
   directories:
   - node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,37 +1,28 @@
 language: node_js
-
 node_js:
-  - "6"
-
+- '6'
 branches:
   only:
-    - master
-
+  - master
 script:
-  - npm run lint:fix
-  - npm run test:cov
-
+- npm run lint:fix
+- npm run test:cov
 cache:
   directories:
-    - node_modules
-
+  - node_modules
 addons:
   code_climate:
-    repo_token: $CODE_CLIMATE
-
+    repo_token: "$CODE_CLIMATE"
 before_install:
-  - "npm install react firebase lodash redux react-redux"
-
+- npm install react firebase lodash redux react-redux
 after_success:
-  - npm install -g codeclimate-test-reporter
-  - codeclimate-test-reporter < coverage/lcov.info
-  - npm run codecov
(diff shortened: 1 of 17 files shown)


prescottprue:
Thanks for the PR!

I would like to get tests passing, and have a little while to test.

Looking into why the tests didn't pass now.

prescottprue:
It looks like the error in the build [might be caused](http://stackoverflow.com/questions/41214625/firebase-node-js-error-the-xmlhttprequest-compatibility-library-was-not-foun) by the new pattern of requiring you used.

We might have to include the whole Firebase library in a top level `require` first.

I am going to look into other ways of making things compatible with the mentioned boilerplate. Might also look into adding an example that uses mentioned boilerplate.

prescottprue:
[v1.1.4](https://github.com/prescottprue/react-redux-firebase/releases/tag/v1.1.4) handles populating lists by `$uid: true`. Will continue to look into addressing boilerplate integration.

michelalbers:
Thank you so much for your work - If I can be of any help please let me know!

prescottprue:
@michelalbers Wanted to ask, are you using yarn or are you disabling it with `USE_YARN=false npm run setup` as listed in [the boilerplate quickstart](https://github.com/mxstbr/react-boilerplate#quick-start)?

prescottprue:
Functionality captured in #53

projoneftw:
Hi, I'm trying to integrate this library with my existing project, based on react-boilerplate. 
I added the following to my enhancers
```
reactReduxFirebase(config, {
      userProfile: 'users',
}
```
The auth and signup is working well, it's saving user profile to Firebase, but it won't load it back.

prescottprue:
@projoneftw Not sure what you mean by "didn't load it back". Are you using `pathToJS(state.firebase, 'profile')` within the `connect` function of the component in which you would like to load the profile?

I think this is more an issue with app code then compatibility with `react-boilerplate` since errors were being thrown before this fix (not just data not loading). Feel free to reach out [over gitter](https://gitter.im/redux-firebase/Lobby) for help confirming.

PR #58, opened 2017-02-05
Add predicate for react-auth-wrapper example

tushardhoot:
Needed for react-auth-wrapper to work properly with react-redux-firebase.


prescottprue:
Great thanks. It has already been updated in the docs too.

PR #74, opened 2017-03-10
confirmPasswordReset action

rpeterson:
* Adds `confirmPasswordReset` action
* Adds tests for `confirmPasswordReset` action
* Fixes typo in `test` npm script

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "clean": "rimraf dist",
     "lint": "eslint src/** test/**",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha -R spec --compilers js:babel-core/register ./tests/setup.js ./test/**/*.spec.js",
+    "test": "mocha -R spec --compilers js:babel-core/register ./tests/setup.js ./tests/**/*.spec.js",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "clean": "rimraf dist",
     "lint": "eslint src/** test/**",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha -R spec --compilers js:babel-core/register ./tests/setup.js ./test/**/*.spec.js",
+    "test": "mocha -R spec --compilers js:babel-core/register ./tests/setup.js ./tests/**/*.spec.js",
```
prescottprue: Great catch! Can't believe this never got switched. I always use `test:cov`, so didn't notice.

src/actions/auth.js
```diff
@@ -388,6 +388,46 @@ export const resetPassword = (dispatch, firebase, email) => {
     })
 }
 
+/**
+ * @description Confirm the password reset with code and password
```
prescottprue: Great job sticking to JSDoc format comments!

prescottprue:
Will be released in `v1.3.2` shortly.

PR #78, opened 2017-03-14
docs(example): remove refs from storage example

C4rmond4i:
I deleted the line `const {newTodo} = this.refs` of the file string upload example because it seemed unnecessary to me.


prescottprue:
Thanks! Was the rest of the example clear other than that?

C4rmond4i:
No problem :)
Yes, simple and easy to understand 👍

Issue #80, opened 2017-03-15
I.hasOwnProperty("process")


darioielardi:
just import reactReduxFirebase and firebaseStateReducer, set them and:
undefined is not a function (evaluating 'I.hasOwnProperty("process")')

prescottprue:
Set them how? Can you provide a repository to reproduce the error? This is one I haven't seen.

How are you "setting them"? In what context?

darioielardi:
1. I install the module, import firebaseStateReducer
2. I declare the firebaseStateReducer as firebase in the combineReducers function
3. I import firebaseConnect
4. The app on the android simulator crash ( red screen ) and the error appears

prescottprue:
Tons of questions:
* Which simulator?
* Is this a `react-native` app?
* Are you using a starter/generator of some kind?
* Are you using webpack/babel or any other type of bundling/building tools?
* If using webpack, what version (i.e. v1 or v2)?

darioielardi:
Sorry for not being​ precise in the first message and thanks for your
answer.
I'm developing on windows 10 a react native app, with the native starter
kit.
When I 'react-native run-android' ( packaged with the react-native packager
) either on the android studio simulator and on the device, the app crash
and that error appears.

Il mer 15 mar 2017, 19:00 Scott <notifications@github.com> ha scritto:

prescottprue:
@darioielardi React-native is not yet supported fully by `react-redux-firebase`. We are working on react-native compatibility on the `react-native` branch.

No promises here, but feel free to checkout the [`react-native` branch](https://github.com/prescottprue/react-redux-firebase/tree/react-native) and [it's associated PR](https://github.com/prescottprue/react-redux-firebase/pull/81). You can even point to within your package file (since it include the dist folder in git tracking) using `git://github.com/prescottprue/react-redux-firebase.git#react-native`.

This functionality will make it back into master and a released version after it is all working correctly so keep an [eye on the pull request](https://github.com/prescottprue/react-redux-firebase/pull/81).

darioielardi:
ok prescott, thanks for the availability.

prescottprue:
@darioielardi Have you gotten a chance to try the `react-native` branch yet? I have begun work on `1.4.0` and would love it include the react-native stuff

darioielardi:
I'd like to give it a try. When I'll do that I will let you know.

prescottprue:
[...]

Issue #53, opened 2017-02-01
Support Boilerplates


prescottprue:
Support the following boilerplates/starters:

* [react-boilerplate](https://github.com/mxstbr/react-boilerplate) - Require sections of Firebase library separately
* [react-cool-starter](https://github.com/wellyshen/react-cool-starter) - Uses module setting from `package.json` instead of loading already babelified version due to usage of Webpack 2

prescottprue:
Versions in the v1.4.0-* range should work:
* react-cool-starter should also work since webpack 2 is supported
* fixes the issue with react-boilerplate by only requiring used parts of Firebase

#### Troubleshooting
Try [release v1.4.0-beta](https://github.com/prescottprue/react-redux-firebase/releases/tag/v1.4.0-beta) first. If this does not work, please post in this issue, and try [release v1.4.0-alpha](https://github.com/prescottprue/react-redux-firebase/releases/tag/v1.4.0-alpha).

There was a slight change in `v1.4.0-beta` which could cause some issues (though it hasn't in any of the boilerplates that have been tested so far).

PR #81, opened 2017-03-15
feat(react-native): react-native and boilerplates support

prescottprue:
* Initial support for `react-native` (should solve #80)
* Import only needed sections of Firebase library ( #72 and #53 )

## Notes About react-native
Support for `react-native` is still in the early testing phases. That means the API and/or functionality may change.

If using `react-native`, make sure you have the following config enabled when creating your store:

```js
import ReactNative from 'react-native'
import { compose, createStore } from 'redux';
import { reactReduxFirebase } from 'react-redux-firebase'
const firebaseConfig = {} // your firebase config object
const initialState = { firebase: { authError: null } };

// then in your config when creating store
const store = createStore(
    makeRootReducer(), 
    initialState,
    compose(
      reactReduxFirebase(
        firebaseConfig,
        {
           rn: ReactNative,
           enableRedirectHandling: false,
           userProfile: 'users' 
         },
      ),
    )
  );
```

Normal authentication methods, such as `firebase.login({provider: 'google', type: 'popup'})`, will not work due to Firebase's js library (used internally) not completely supporting `react-native`. That means you must use `firebase.auth().signInWithCredential(credential)` until there is more support within the Firebase library.

It might look something like this (untested):

```js
let credential = this.props.firebase.auth.GoogleAuthProvider.credential(token);
this.props.firebase.auth().signInWithCredential(credential)
  .then(() => {
    console.log('auth successful')
  })
  .catch((err) => {
    console.log('error: ', err)
  })
```


prescottprue:
`react-native` branch will remain until `v1.4.0` full version is released (not preversions such as `v1.4.0-alpha`

PR #88, opened 2017-03-22
Return file snapshot on upload

urbantumbleweed:
#### What's this PR do?

The promise from 'uploadFile' now returns an object of the file data is decorated with a `pushKey` property that is the key generated by the call to `firebase.push`.  This change will give the caller the ability to access the urls, name and key for additional processing.

#### What are the important parts of the code?
a single `.then` with a spread

#### How should this be tested by the reviewer?
Ensure the caller of this function has access to the fileData in the returned promise.

#### Is any other information necessary to understand this?
This change is proposed as the caller may need to dispatch additional actions for use with redux/redux-observable that contain the information contained in the fileData.

### DIFF
--- a/src/actions/storage.js
+++ b/src/actions/storage.js
@@ -81,7 +81,7 @@ export const uploadFile = (dispatch, firebase, { path, file, dbPath }) =>
       return firebase.database()
         .ref(dbPath)
         .push(fileData)
-        .then(fileSnap => ({ pushKey: fileSnap.key, ...fileData }))
+        .then(fileSnap => ({ ...fileSnap, File: fileData }))
     })
 
 /**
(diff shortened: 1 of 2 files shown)


src/actions/storage.js
```diff
@@ -81,6 +81,7 @@ export const uploadFile = (dispatch, firebase, { path, file, dbPath }) =>
       return firebase.database()
         .ref(dbPath)
         .push(fileData)
+        .then(fileSnap => ({ pushKey: fileSnap.key, ...fileData }))
```
prescottprue: @urbantumbleweed I like the idea of having the whole snapshot available there instead of just the key under a different name.

What are your thoughts on `.then(fileSnap => ({ ...fileSnap, File: fileData }))`

This way, you can access like so:
```
uploadFile()
   .then((snapWithFile) => {
     console.log('key:', snapWithFile.key)
     console.log('file data:', snapWithFile.File) // which shouldn't run over any snapshot paramters
   })
```

Issue #92, opened 2017-03-24
equalTo, startAt and endAt parses every string starting with digit


biomorgoth:
I am trying to run a query that consists of email addresses that starts with digit. Let's use 123example@gmail.com as an example. When i try to do the following:

```
firebaseConnect([ 'todos#orderByChild=emailAddress&equalTo=123example@gmail.com' ])(SomeComponent);
```

It does not return the same matches as i built the query myself:

```
getFirebase().database().ref('todos').orderByChild('emailAddress').equalTo('123example@gmail.com')
```

And researching into the code, i noticed in method `applyParamsToQuery ` from `utils/query.js`, that when case is `equalTo`, `startAt` and `endAt`, a `parseInt` is applied on the param value. This makes react-redux-firebase build a different query that equals to this:

```
getFirebase().database().ref('todos').orderByChild('emailAddress').equalTo(123)
```

Is there a way to circumvent this situation? I cannot see a way to avoid it without making changes to that method.

prescottprue:
@biomorgoth Your syntax is correct and should work. This seems to be a bug within the exact method that you pointed out.

I will begin work on solving it an post updates here. Thanks for reporting!

biomorgoth:
@prescottprue As a fast and non-invasive solution to the issue, i thought of adding a `equalToStr` case in `applyParamsToQuery` that explicitly skips the `parseInt` call, avoiding all the situation. But i do not know if you prefer to keep all the same names just like Firebase.

A bigger approach would be providing an object `queryParameters` to `applyParamsToQuery` instead of an array, so we can specifically pass Integer/Float/String/etc. values without parsing it internally. But that still leaves the query string in `firebaseConnect` broken, if anyone would want to use it (and i personally like it a lot!).

prescottprue:
I think it might be good to better test for if that is a number before parsing. We should keep all the existing functionality (string or object notation).

Either way, starting to look into solutions now.

biomorgoth:
In fact i have just came with a possible solution by using `Number(value)` instead of `parseInt(value)`, which gives the following advantages:

- Tests is value is completely a number or returns NaN.
- Returns an Integer or Float result depending of parsed value. Note that with current `parseInt` doing `equalTo(8.9)` is not possible either, as it is actually translating to `equalTo(8)`.

I am testing that solution and will make a pull request if everything is fine.

prescottprue:
Awesome, thanks!

prescottprue:
[...]

PR #94, opened 2017-03-24
Fix applyParamsToQuery: make equalTo, startAt and endAt parse numbers…

biomorgoth:
… correctly if needed.

Solves issue #92

### DIFF
--- a/src/utils/query.js
+++ b/src/utils/query.js
@@ -2,6 +2,14 @@ import { actionTypes } from '../constants'
 
 const { UNSET_LISTENER } = actionTypes
 
+const tryParseToNumber = (value) => {
+  const result = Number(value)
+  if (Number.isNaN(result)) {
+    return value
+  }
+  return result
+}
+
 /**
  * @private
  * @description Get path to watch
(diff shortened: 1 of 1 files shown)


src/utils/query.js
```diff
@@ -2,6 +2,14 @@ import { actionTypes } from '../constants'
 
 const { UNSET_LISTENER } = actionTypes
 
+const tryParseToNumber = (value) => {
+  const result = Number(value)
```
prescottprue: This doesn't ever throw does it? even if value is undefined?

biomorgoth: According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number):

> If the argument cannot be converted into a number, it returns NaN.

And effectively, i've just tested in Chrome, Firefox and Node and `Number(undefined)` returns `NaN`. So, `tryParseToNumber` would return the undefined itself.

prescottprue:
Great fix!

Planning on releasing this in a patch sometime today (hopefully after adding to the tests to cover this). Will post here on release even though issue will be closed.

PR #112, opened 2017-04-12
fix(docs): fixed authenticatingSelector documentation

emilguareno:
The isInitializing selector should be the same regardless of the predicate


prescottprue:
Great find.

Thanks for the pull request!

PR #117, opened 2017-04-27
Improve non-decorator connect example

timkindberg:


prescottprue:
Will publish this to the docs site in the next day or so.

PR #122, opened 2017-05-04
Use prop-types package instead of React.PropTypes

petetnt:
Signed-off-by: petetnt <pete.a.nykanen@gmail.com>

### Description

The `React.PropTypes` have been deprecated and moved to `prop-types` package. Replaces instances of `React.PropTypes` with the `prop-types` package to remove deprecation warnings from current users.

For more information see https://facebook.github.io/react/warnings/dont-call-proptypes.html

### Check List

- [x] All test passed


prescottprue:
Great catch!

Planned to be released as part of `v1.4.0-rc.2`.

PR #123, opened 2017-05-05
Population of profile params

fej-snikduj:
### Description
Non-breaking small feature addition to profile parameter population.  Profile population will now work on a list of UIDs, key:value pairs, and a single UID (which was already supported).

### Check List

- [X] All test passed
- [ ] Added test to ensure to fix/ensure properly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "build:umd": "cross-env BABEL_ENV=commonjs NODE_ENV=development webpack src/index.js dist/react-redux-firebase.js",
     "build:umd:min": "cross-env BABEL_ENV=commonjs NODE_ENV=production webpack src/index.js dist/react-redux-firebase.min.js",
     "build": "npm run build:commonjs && npm run build:es && npm run build:umd && npm run build:umd:min",
-    "watch": "npm run build:commonjs -- --watch",
+    "watch": "npm run build:umd -- --watch",
     "prepublish": "npm run clean && npm run build",
     "prepush": "npm run lint:fix",
     "docs:clean": "rimraf _book",
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -6,11 +6,16 @@ import {
   forEach,
   set,
   get,
+  map,
   mapValues
 } from 'lodash'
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -18,7 +18,7 @@
     "build:umd": "cross-env BABEL_ENV=commonjs NODE_ENV=development webpack src/index.js dist/react-redux-firebase.js",
     "build:umd:min": "cross-env BABEL_ENV=commonjs NODE_ENV=production webpack src/index.js dist/react-redux-firebase.min.js",
     "build": "npm run build:commonjs && npm run build:es && npm run build:umd && npm run build:umd:min",
-    "watch": "npm run build:commonjs -- --watch",
+    "watch": "npm run build:umd -- --watch",
```
prescottprue: Nice. This is the more often use-case since most are requiring and using Webpack or some other bundler that points to the UMD version.

src/actions/auth.js
```diff
@@ -105,7 +119,39 @@ export const watchUserProfile = (dispatch, firebase) => {
                 const populates = getPopulateObjs(profileParamsToPopulate)
                 const profile = snap.val()
                 forEach(populates, (p) => {
-                  set(profile, p.child, get(data, `${p.root}.${snap.val()[p.child]}`))
+                  const child = get(profile, p.child)
```
prescottprue: Nice use of get here, this will keep it so that dot notation can be used for child parameter of populate objects.

src/utils/populate.js
```diff
@@ -25,6 +25,24 @@ export const getPopulateObj = (str) => {
 }
 /**
  * @private
+ * @description Determine the structure of the child parameter to populate onto
+ * @param {String|Object} child - Value at child parameter
+ */
+export const getChildType = (child) => {
```
prescottprue: This is great for now, but eventually we might want to make this more general (return type of all types of vars) to be used elsewhere. Wish there was a lodash function to just return a variable's type [but it seems like they feel as though developers should make this themselves](https://github.com/lodash/lodash/issues/391).

Issue #130, opened 2017-05-13
bug(populates): populates clears all data at root


Domiii:
The following query causes all other data currently stored at `users` (which is the `root`) to be deleted. As soon as the `populates` portion is resolved, all my (self-managed) users are gone from my client (however the `users` path is still in the `firebaseConnect` mix):

```js
populates: [
  { child: 'uid', root: 'users', keyProp: 'uid' }
]
```

**UPDATE**

At some point, [this SET action](https://github.com/prescottprue/react-redux-firebase/blob/master/src/actions/query.js#L146-L153) is dispatched with `path` = `users`:

```js
dispatch({
  type: SET,
  path,
  data: result,
  timestamp: Date.now(),
  requesting: false,
  requested: true
})
```

From the looks of it, that is because [promisesForPopulate](https://github.com/prescottprue/react-redux-firebase/blob/master/src/utils/populate.js#L112) actually creates a big object with all the `user` children which then overrides all previous data at `root` (= `/users`), using the above `SET` action.

### Possible Fix (1)

Each fetched result object should be set with it's own `SET` action. So instead of:

```js
SET('/users', {
  uid1: {...},
  uid2: {...}
});
```

you get:

```js
SET('/users/uid1', {...});
SET('/users/uid2', {...});
```

### Possible Fix (2)

Add an `ADD` or `UPDATE` action which adds to rather than overrides the given data at the given `root` path:

```js
ADD('/users', {
  uid1: {...},
  uid2: {...}
});
```

prescottprue:
Have you tried using storeAs for your other users query?

Domiii:
Thanks! I'll try that as a work-around.

When do you think `populate` will be fixed though?

prescottprue:
@Domiii placing the populate results in a different place in redux would also require modifying `populatedDataToJS`. Data is placed  into redux under its root.

Domiii:
I tried to give it a swing, but `NPM` bugs keep holding me back.

I added my own fork of `react-redux-firebase` as a local git submodule which is a bit of a pain (see below).

### PS: (For those who are interested) How to set this up as an NPM submodule?

NOTE: I know this does not technically belong into this thread. If you don't need/want it, I'll delete it.

WARNING #1: Git does not make it easy to remove or change submodules, gotta move through a few steps there, but you can google that easily.

WARNING #2: NPM does not handle local sub modules well. Here is how I got it to work:

[...]

PR #132, opened 2017-05-15
Adds various features to populates

javamonn:
### Description

Adds various features supporting `populates` that we found to be missing for our use-case. I've ensured the tests pass and functionally everything appears to work. We could consider splitting this PR up if some features are deemed undesired for merge.

- Adds support for lodash supported path syntax in `populates.child`, e.g. `this.is.a.nested.prop`. This was added to support populating on ids that may be nested arbitrarily in the top level object and primarily required changing direct JS prop accessors `foo[bar], foo.bar` to `_.get(foo, bar)`
- Adds support for per-item populates configurations by specifying a function callback that generates the populates configuration from the top level (key, item) tuple. This enables dynamically populating items, such as using the top level key when building the `root` path, or to only populate items passing a test when querying for a list.
- Adds support for `storeAs` when `populates` is configured to continue working around https://github.com/prescottprue/react-redux-firebase/issues/130 - is there a reason why this shouldn't work?

### Check List

- [x] All test passed
- [x] Added test to ensure to fix/ensure properly.


javamonn:
Hmm, the test suite passes for me locally, will have a look later this afternoon.

prescottprue:
These are awesome additions to populate! Going to test it out today, and look into add some stuff to the docs.

prescottprue:
These features have been [added to the roadmap for `v1.5.0`](http://react-redux-firebase.com/docs/roadmap.html#next-minor-version-v150).

Going to mere into `v1.5.0` branch for now. Hoping to have an alpha version of `v1.5.0` soon.

src/actions/query.js
```diff
@@ -148,7 +149,7 @@ export const watchEvent = (firebase, dispatch, { type, path, populates, queryPar
           })
           dispatch({
```
prescottprue: This got moved to the dispatch after the populate children dispatches (keeps `isLoaded` from being `true` before all children are set to redux)

Issue #29, opened 2017-01-11
feat(firebase): multiple firebase apps


rafaelcorreiapoli:
Is it possible to connect with multiple firebase apps?
My current use case is to connect to one central firebase app that has all the keys of other firebase apps. Once the user has logged in to this app, it retrieves the key for the corresponding app and connects to it

prescottprue:
@rafaelcorreiapoli It isn't currently supported, but I will work on getting included in v1.2.0.

Do you have a prefered API for this?

I was thinking something like:

```js
const apps = {
  app1: {
   apiKey: '<your-api-key>',
   authDomain: '<your-auth-domain>',
   databaseURL: '<your-database-url>',
   storageBucket: '<your-storage-bucket>'
  },
  app2: {
   apiKey: '<your-api-key>',
   authDomain: '<your-auth-domain>',
   databaseURL: '<your-database-url>',
   storageBucket: '<your-storage-bucket>'
  }
}
const 
// React Redux Firebase Config
const config = {
  userProfile: 'users'
}

// Add react-redux-firebase to compose
const createStoreWithFirebase = compose(
 reactReduxFirebase({ apps }, config),
)(createStore)
```

Maybe it shouldn't be named "apps", but it would be nice to have it be differentiated from the normal object [passed for a single app](http://react-redux-firebase.com/docs/api/compose.html).

rafaelcorreiapoli:
Hi! Thanks for the answer
It actually would not fix my problem
I need to instantiate my second app based on informations I get after login in to the first app
The first app holds the information about other apps (apiKey, authDomain, etc...)
I know its not a common use case.. but do you have any ideas on how to solve this ?

prescottprue:
OH man... Yeah that changes things quite a bit. There is for sure not a way to do that in an easy way currently, especially because all of the data syncing and other methods use the default app.

I will add this to the roadmap so that it is on the menu at least, but it will take quite a bit of research to see how all the methods would need to be modified.

You might want to look into [re-base](https://github.com/tylermcginnis/re-base) for a more simple integration that easily allows for multiple apps. Even though it only syncs to components (not redux) it should handle your use case.

rafaelcorreiapoli:
Nice! thanks! Maybe I could render another component tree with a different Provider when I get the data from the first app. This new Provider would use a store created with data from the first app
```
  const createStoreWithFirebase = compose(
    reactReduxFirebase(firebaseConfig, { userProfile: 'users' }),
[...]

PR #148, opened 2017-05-31
Allow to use different stores in firebaseConnect

martynovs:
### Description
It allow to use more than one firebase app by using separate store for each one and connect components to different stores.

### Questions
- Does this impact the external API? **YES**

### Check List
- [-] All tests passing
- [-] Docs updated with any changes or examples
- [-] Added tests to ensure feature(s) work properly

### Relevant Issues
* [#29](https://github.com/prescottprue/react-redux-firebase/issues/29)

### DIFF
--- a/src/connect.js
+++ b/src/connect.js
@@ -31,7 +31,7 @@ import { getEventsFromInput, createCallable } from './utils'
  *   auth: pathToJS(firebase, 'auth') // pass auth data as this.props.auth
  * }))(fbWrapped)
  */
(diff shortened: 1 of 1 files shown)


prescottprue:
Interesting. Is there a reason that having two different stores is preferred to running two firebase reducers within one store?

martynovs:
My app works for many customers, each customer have separate database. Admins can switch between customers.
I want to use default store for application wide components, and separate store for customer related components. When admin switch customer I can drop store and create new one for next customer db.

PR #149, opened 2017-05-31
Don't mix import/export with module.exports

martynovs:
### Description
Webpack 2 doesn't allow to mix import/export with module.exports.
This forbids to use src in webpack, which is required for dead code elimination (tree shaking) to work.

### Questions
- Does this impact the external API? **NO**

### Check List
- [ ] All tests passing

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -156,13 +156,3 @@ export default {
   metaParams,
   paramSplitChar
 }
-
-module.exports = {
-  defaultJWTProps,
-  actionTypes,
-  defaultConfig,
-  supportedAuthProviders,
-  defaultInitProps,
-  metaParams,
-  paramSplitChar
-}


prescottprue:
Fix should be good, testing a little now.

Question: How are you importing `src` through externally though webpack? The `jsnext:main` and `module` fields in the package file point to the es5ified version, so is it not possible to just use that?

martynovs:
```js
import { reactReduxFirebase } from 'react-redux-firebase/src'
```
I import this way because webpack require es6 code with import/export for tree shaking to work.
All libraries that I use point their `module` or `jsnext:main` to es6 sources and webpack uses it, but `react-redux-firebase/es` contains code processed by babel so there is no import/export.

prescottprue:
@martynovs From what I have seen ([as it is in redux](https://github.com/reactjs/redux/blob/master/package.json#L8) it is common to do this.

You importing or requiring this in your application shouldn't have to do with whether it is in es5 or es6 right? You don't use the import/export from `react-redux-firebase`?

martynovs:
For es build it doesn't transform import/export to common-js modules https://github.com/reactjs/redux/blob/master/.babelrc#L27

prescottprue:
@martynovs The link you provided points to the settings for the commonjs bundle instead of the es bundle (which does not do that).

Would you be open to chatting over [gitter](https://gitter.im/redux-firebase/Lobby)?

martynovs:
It's late night for me. This simple change just removes code duplication and fixes mixing issue.

prescottprue:
@martynovs Will merge to v1.5.0, will have to be released in beta.2.

I would still like to discuss tree-shaking since it would be good to keep users from needing to point to src (again since then they would need to deal with all of the babel transforms of the library within their application).

PR #150, opened 2017-05-31
Extract firebase helpers initialization

martynovs:
### Description

Extract initialization to separate module will allow to configure more than one firebase app and use externally loaded firebase (from firebase hosting, not from npm).
Also library will not dictate to use it's `compose`, in my case I use `compose` from redux dev tools.

store configuration may look like this:
```js
import firebaseInit from 'react-redux-firebase/src/init';

function attachFirebaseToStore(store, config, name) {
  const app = firebase.initializeApp(config, name);
  const instance = firebaseInit(app, config, store.dispatch);
  store.firebase = instance;
  return instance;
}
```

### Questions
- Does this impact the external API? **YES**

### Check List
- [ ] All tests passing


prescottprue:
Nice work! I have been wanting to do this for a while now. Going to test it out.

src/init.js
```diff
@@ -0,0 +1,346 @@
+import { isObject } from 'lodash'
+import { authActions, queryActions, storageActions } from './actions'
+
+export default function init (firebase, config, dispatch) {
```
prescottprue: It would be nice to see this as a `const` set to a function instead of a named function to limit scope unless there is a reason it is being named:

```js
export default (firebase, config, dispatch) => {
```

src/init.js
```diff
@@ -0,0 +1,346 @@
+import { isObject } from 'lodash'
```
prescottprue: It might help for clarity to call this file "createFirebaseInstance" or something along those lines.

martynovs:
I've applied your changes

prescottprue:
I believe the branch was made from master instead of v1.5.0. 

Going to make similar changes and attempt to resolve conflicts.

prescottprue:
Similar has been done and released in [`v2.0.0-alpha`](https://github.com/prescottprue/react-redux-firebase/releases/tag/v2.0.0-alpha).

Hopefully will soon get time to try to get it into a `v1.*.*` version.

PR #156, opened 2017-06-10
docs(connect): Expand on argument documentation for `firebaseConnect`

JustinTulloss:
### Description

The existing docs didn't make it clear that you could have paths that
depended on props. I expanded on the argument documentation and added an
example.

I also regenerated the docs to get the changes into the actual markdown files, but it looks like there are quite a few changes that haven't been committed.

### Questions

- Will a new version need to be released or is this a docs change? docs change
  - Which version should this be published as a part of? no idea

### Check List

- [x] All tests passing
- [x] Docs updated with any changes or examples
- [x] Added tests to ensure feature(s) work properly


prescottprue:
@JustinTulloss Part of what you are seeing with the changes that aren't committed is intentional (it is a duplication of docs in two places). In the future (#150 starts to capture this) that section will be in a separate file so this duplication and not committing of files will go away.

I will handle the removal of stuff that shouldn't be in there this time since it will be fixed soon.

Thanks for the PR!

JustinTulloss:
@prescottprue awesome, thanks!

PR #170, opened 2017-06-24
[v2.0.0] fix(initialize): Check that 'database' is a function instead of 'initializeApp'

agentilela:
Check that 'database' is a function instead of 'initializeApp'. This allows for passing in of pre-initialized firebase instances (like react-native-firebase).

### Description

When using react-native-firebase:

```
import firebase from 'react-native-firebase'
import { reactReduxFirebase } from 'react-redux-firebase'

const FBConfig = {
  debug: false,
  persistence: true
}

const reactReduxFBConfig = {
  enableLogging: false,
  userProfile: 'users',
}

const firebaseApp = firebase.initializeApp(FBConfig)

compose(
reactReduxFirebase(firebaseApp, reactReduxFBConfig)
...
)
```

Returns the error: **databaseURL is a required config parameter for react-redux-firebase**

In this case, react-redux-firebase tries to parse firebaseApp as a config object, because it no longer contains the initalizeApp function. This PR addresses that. 

One concern I still have though- I am not sure whether the database() function exists on firebase instances with no database configured. So there may be a better way to address this.

### Check List

- [ ] All tests passing

### DIFF
--- a/src/compose.js
+++ b/src/compose.js
@@ -91,7 +91,7 @@ export default (fbConfig, otherConfig) => next =>
     let firebaseInstance
 
     // handle firebase instance being passed in as first argument
-    if (typeof fbConfig.initializeApp === 'function') {
+    if (typeof fbConfig.database === 'function') {
       firebaseInstance = createFirebaseInstance(fbConfig, otherConfig, dispatch)
     } else {
       // Combine all configs


prescottprue:
Released as part of [`v2.0.0-alpha.2`](https://github.com/prescottprue/react-redux-firebase/releases/tag/v2.0.0-alpha.2)

PR #185, opened 2017-07-07
Fix issue with signInWithCredential

buraks:
[Work In Progress]

`firebase.auth().signInWithCredential` is expecting a `firebase.auth.AuthCredential` object (https://firebase.google.com/docs/reference/js/firebase.auth.Auth#signInWithCredential)

### DIFF
--- a/src/utils/auth.js
+++ b/src/utils/auth.js
@@ -45,6 +45,17 @@ export const createAuthProvider = (firebase, providerName, scopes) => {
   return provider
 }
 
+const _createAuthCredential = (firebase, providerName, token) => {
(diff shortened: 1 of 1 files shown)


buraks:
this solution would work for `facebook`, `github` and `google`. However, as `twitter` needs `secret` as well as `token` (https://firebase.google.com/docs/reference/js/firebase.auth.TwitterAuthProvider#.credential) this solution is not enough.

maybe this `getLoginMethodAndParams` can simply accept `credential` as a parameter so that the logic can be upto the caller. what do you think?

prescottprue:
@buraks Checkout #186

Added support for passing credential. Got rid of support for token and provider together since the providers have different settings.

PR #187, opened 2017-07-07
Fix createUserProfile on login with Facebook

buraks:
When a user logs in with Facebook, the userData is in the following format:

```javascript
{
  email: ...
  displayName: ...
  providerData: [
    {
      email: ...
      photoURL: ...
      providerId: facebook.com
      uid: ...
      displayName: ...
      ...
    }
  ]
  ...
}
```

there is no `user` in `userData` so it was failing creating the profile. This pr fixes that.

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -90,8 +90,8 @@ export const watchUserProfile = (dispatch, firebase) => {
 }
 
 /**
- * @description Create user profile if it does not already exist. `updateProifleOnLogin: false`
- * can be passed to config to dsiable updating. Profile factory is applied if it exists and is a function.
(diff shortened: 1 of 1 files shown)


src/actions/auth.js
```diff
@@ -260,7 +260,7 @@ export const login = (dispatch, firebase, credentials) => {
       if (!userData) return Promise.resolve(null)
 
       // For email auth return uid (createUser is used for creating a profile)
-      if (userData.email) return userData.uid
+      if (method === 'signInWithEmailAndPassword') return userData.uid
```
prescottprue: Nice! Exactly what I was thinking.

PR #188, opened 2017-07-09
feat(firebaseConnect): expose wrapped component and displayName

JustinTulloss:
### Description

Adds two statics to a connected component:
- `WrappedComponent`, which is the original component. This is useful for testing without mocking out firebase.
- `displayName` is set to reveal what component is wrapped by FirebaseConnect. This makes debugging easier.

This will require a new version (presumably 1.4.4) but is purely additive and therefore should not break any existing code, unless people were relying on the previous behavior of `displayName`.

### Check List

- [x] All tests passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "immutable": "^3.8.1",
     "jwt-decode": "^2.2.0",
     "lodash": "^4.17.4",
-    "opencollective": "^1.0.3"
+    "opencollective": "^1.0.3",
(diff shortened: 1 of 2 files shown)


prescottprue:
Great addition. Will release in `v1.4.4` soon.

prescottprue:
Released with matching tests in [`v1.4.4`](https://github.com/prescottprue/react-redux-firebase/releases/tag/v1.4.4).

PR #213, opened 2017-07-29
make notSetValue optional in typings

flash1293:
### Description

This PR modifies the typings and makes the fallback-value (`notSetValue`) optional in all methods. This makes it possible to use it with two parameters and the implicit fallback-value of `undefined`, e.g.  `dataToJS(firebase, 'todos')` instead of `dataToJS(firebase, 'todos', undefined)`.

The docs don't have to be changed as this usage is intended and already possible in plain JS. The example above is copied from the docs.

Because this API-change is backwards compatible, no major release is necessary. I recommend to include this into the upcoming `1.5.0` release.

### Check List

- [ ] All tests passing - Tests are failing with `Error: Cannot find module 'react-display-name'`. However I suspect this is a separate issue (also happens with the upstream-branch)
- [x] Docs updated with any changes or examples
- [x] Added tests to ensure feature(s) work properly

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -105,9 +105,9 @@ export interface listenerConfigFunc {
 
 export function buildChildList(data: any, list: any, p: any): any;
 
-export function customToJS(data: any, path: any, custom: any, notSetValue: any): any;
+export function customToJS(data: any, path: any, custom: any, notSetValue?: any): any;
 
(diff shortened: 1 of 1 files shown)


flash1293:
Was pointed to the wrong upstream, sorry. Opened a new one.

PR #214, opened 2017-07-29
make notSetValue parameter optional in typings

flash1293:
### Description

This PR modifies the typings and makes the fallback-value (`notSetValue`) optional in all methods. This makes it possible to use it with two parameters and the implicit fallback-value of `undefined`, e.g.  `dataToJS(firebase, 'todos')` instead of `dataToJS(firebase, 'todos', undefined)`.

The docs don't have to be changed as this usage is intended and already possible in plain JS. The example above is copied from the docs.

Because this API-change is backwards compatible, no major release is necessary. I recommend to include this into the upcoming `1.5.0` release.

### Check List

- [x] All tests passing
- [x] Docs updated with any changes or examples (doesn't apply)
- [x] Added tests to ensure feature(s) work properly (doesn't apply)


prescottprue:
[v1.5.0-rc.3](https://github.com/prescottprue/react-redux-firebase/releases/tag/v1.5.0-rc.3) includes this fix

PR #221, opened 2017-08-06
fix(listeners): deep path listeners not working bug (#219)

### LINKED ISSUE #219: [v2.0.0] bug(listeners): deep path listeners not working
Bug

When subscribing to multiple paths from Firebase, each reactReduxFirebase/SET action overwrites the firebase.data state entirely, wiping away any previous state from any other subscriptions.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via https://jsfiddle.net or similar.**
Subscribe to two paths with firebaseConnect and view the state at each reactReduxFirebase/SET action.

**What is the expected behavior?**
The data should be merged with the previous state.

I noticed this when migrating my working 1.5 code to 2.0.0-beta.3.

marekolszewski:
### Description
Fixes #219 by using lodash's `set` one level higher in the reducer. 

### Check List

- [X] All tests passing
- [N/A] Docs updated with any changes or examples
- [N/A] Added tests to ensure feature(s) work properly

### Relevant Issues
#219

### DIFF
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -149,15 +149,9 @@ export const timestampsReducer = (state = {}, { type, path }) => {
 const createDataReducer = (actionKey = 'data') => (state = {}, action) => {
   switch (action.type) {
     case SET:
-      return {
-        ...state,
-        ...set({}, getDotStrPath(action.path), action[actionKey])
-      }
+      return set(state, getDotStrPath(action.path), action[actionKey])
     case NO_VALUE:
-      return {
-        ...state,
-        ...set({}, getDotStrPath(action.path), null)
-      }
+      return set(state, getDotStrPath(action.path), null)
     case LOGOUT:
       // support keeping data when logging out - #125
       if (action.preserve) {
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -1,6 +1,7 @@
-import { combineReducers } from 'redux'
-import { set, pick, omit } from 'lodash'
+import set from 'lodash/fp/set'
 import { actionTypes } from './constants'
+import { combineReducers } from 'redux'
+import { pick, omit } from 'lodash'
 
 const {
   START,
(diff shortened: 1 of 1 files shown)


prescottprue:
@marekolszewski I believe that this is an anti pattern of redux due to the not keeping [to the immutablility pattern](http://redux.js.org/docs/faq/ImmutableData.html#why-is-immutability-required).

The reason for the spread operator is to create a copy of state (is transform to the same as using `Object.assign`).

This is the code that is causing the issue, but I am hoping for a way that does not modify state.

marekolszewski:
@prescottprue You're right, sorry, I forgot that set modifies the original object. Let me change it.

marekolszewski:
@prescottprue OK, just updated the PR to use lodash's immutable version of set.

src/reducer.js
```diff
@@ -149,9 +150,9 @@ export const timestampsReducer = (state = {}, { type, path }) => {
 const createDataReducer = (actionKey = 'data') => (state = {}, action) => {
   switch (action.type) {
     case SET:
-      return set(state, getDotStrPath(action.path), action[actionKey])
+      return set(getDotStrPath(action.path), action[actionKey], state)
```
prescottprue: Not sure this is what you want either. This points to the string path as the original data (which will always be a string). This set will need to be part of a spread/assign with state as it exists currently.

marekolszewski: Lodash FP functions have different argument orders from the corresponding non-FP versions (they did this to make composition easier). More details here: https://github.com/lodash/lodash/wiki/FP-Guide#rearranged-arguments

prescottprue: You are totally right! I didn't notice the `fp` in the import.

[This lodash issue confirms this usage](https://github.com/lodash/lodash/issues/1696#issuecomment-269394600). Thanks for the contribution!

PR #233, opened 2017-08-12
Initialize initialState

novito:
### Description

If one is following the Readme it will get an error because initialState doesn't have a value. I just added a empty `initialState`


prescottprue:
Thanks 👍

PR #235, opened 2017-08-14
Correctly unwatch multiple queries to the same base path.

### LINKED ISSUE #234: Does not unwatch all listeners upon unmount if there are multiple listeners with different queries at the same base path.
*bug*

When using storeAs to create to unique queries to the same path, the listeners are not correctly unwatched upon component dismounting.  The initial watching works correctly, and all functions well, until the component dismounts.  Only one of the listeners is unset, and the remaining one remains on.  When the component remounts, the storeAs functionality no longer works as it did upon first mounting (now only one listener is currently on).
````
  @firebaseConnect(({ userId1, userId2 }) => {
    const query1 = [
      'orderByChild=createdBy',
      `equalTo=${userId1}`,
    ];
    const query2 = [
      'orderByChild=assignedTo',
      `equalTo=${userId2}`,
    ];

    return [
      { path: 'todos', queryParams: query1, storeAs: 'user1Todos' },
      { path: 'todos', queryParams: query2, storeAs: 'user2Todos' },
    ];
  }
````

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via [codesandbox](https://codesandbox.io/) or similar.**

**What is the expected behavior?**
Expected behavior is to un-watch both queries upon dismount and to rewatch both upon remount.

v1.4.5

fej-snikduj:
### Description
When multiple listeners are set on the same path with different query parameters, all of the paths are not correctly unwatched when the component unmounts.

Passing queryId into the unWatchEvent function is necessary for
unique queries on the same base path to be properly identified when
unwatching events.  getQueryIdFromPath and getWatchPath do not
correctly identify unique queries at the same path (because the path
that is passed into these functions is the path without query strings
attached).  This scenario happens when using storeAs functionality.

### Check List

- [X ] All tests passing (except one test that was failing prior to this PR)
- [NA] Docs updated with any changes or examples if applicable
- [NA ] Added tests to ensure new feature(s) work properly

### Relevant Issues

Closes #234

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-redux-firebase",
-  "version": "1.4.5",
+  "version": "1.4.6",
   "description": "Redux integration for Firebase. Comes with a Higher Order Component for use with React.",
   "browser": "dist/react-redux-firebase.js",
   "main": "lib/index.js",
--- a/src/actions/query.js
+++ b/src/actions/query.js
@@ -193,7 +193,7 @@ export const watchEvents = (firebase, dispatch, events) =>
  */
 export const unWatchEvents = (firebase, dispatch, events) =>
     events.forEach(event =>
-      unWatchEvent(firebase, dispatch, event.type, event.path)
+      unWatchEvent(firebase, dispatch, event.type, event.path, event.queryId)
     )
 
 export default { watchEvents, unWatchEvents }


prescottprue:
Thanks! Tests don't pass, but it seems like an easy fix that we can handle later.

Will release it tonight.

PR #237, opened 2017-08-15
[v2.0.0] feat(reducers): authError reducer to mirror authError state in v1

marekolszewski:
### Description
V2 appears to have dropped the firebase.authError redux state present in v1.x. This PR adds it back to maintain better backwards compatibility.

### Check List

- [X] All tests passing

### DIFF
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -216,6 +216,27 @@ export const authReducer = (state = { isLoaded: false, isEmpty: true }, action) 
 }
 
 /**
+ * Reducer for authError state. Changed by `LOGIN`, `LOGOUT`, `LOGIN_ERROR`, and
+ * `UNAUTHORIZED_ERROR` actions.
+ * @param  {Object} [state={}] - Current authError redux state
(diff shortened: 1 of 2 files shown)


prescottprue:
@marekolszewski Nice work! Thanks for following the JSDoc comment pattern, that makes keeping docs updated much easier.

Originally leaving `authError` out was intentional so that everything would be handled with the errors reducer, but it makes more sense to keep it more compatible with v2.

nehvaleem:
@prescottprue - could you please tell me when you plan to publish beta.6 to npm? I could use this authError in my project :)

prescottprue:
@nehvaleem Apologies, I was hosting an event over the last view days.

It should be able to be published tonight after I confirm a few things.

prescottprue:
@nehvaleem It has been released with a few docs updates. 

Future releases should not be held up, I'm working with a few people to help lighten the load.

nehvaleem:
@prescottprue awesome! thanks for your support.

Issue #223, opened 2017-08-08
[v2.0.0] bug(populate): deep path roots do not work with population


marekolszewski:
**Do you want to request a *feature* or report a *bug*?**  
Bug

**What is the current behavior?**

When specifying a populates parameter with a multi-level root path, like: `[{ child: 'user', root: 'public/users' }]`, to the the `populate` function, it will silently fail to populate the targeted redux state correctly.

**What is the expected behavior?**
It should populate the child key correctly.

**Which versions of dependencies, and which browser and OS are affected by this issue? Did this work in previous versions or setups?**
I noticed this bug when migrating from v1.5 to v2.0.0-beta. The same code was working correctly with 1.5.

PR #241, opened 2017-08-16
fix(populate): Update to deep path roots do not work with population (v2.0.0)

marekolszewski:
### Description
This is an update to PR 224, which did not go far enough at deeply merging data into the redux state.

Note: lodash's merge function appears to do a deep clone of the merged objects, which is likely unnecessary and expensive here. It would be nice to avoid this but there is no obvious way to accomplish this with lodash so I'm leaving it for a later PR.

### Check List

- [X] All tests passing

### Relevant Issues
#223

### DIFF
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -1,6 +1,6 @@
-import set from 'lodash/fp/set'
 import { actionTypes } from './constants'
 import { pick, omit } from 'lodash'
+import { flow, set, merge } from 'lodash/fp'
 
 const {
   START,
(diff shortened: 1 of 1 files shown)


src/reducer.js
```diff
@@ -172,7 +172,10 @@ export const timestampsReducer = (state = {}, { type, path }) => {
 const createDataReducer = (actionKey = 'data') => (state = {}, action) => {
   switch (action.type) {
     case SET:
-      return set(getDotStrPath(action.path), action[actionKey], state)
```
prescottprue: Is this because set it not copying the whole state object?

marekolszewski: Set overwrites the existing data at the path you are writing to. For example, if you are writing a bunch of new users from /public/users, any previous users in firebase.data.public.users will be overwritten rather than merged with.

PR #297, opened 2017-10-18
fix(populate): multiple populates when missing child key

### LINKED ISSUE #277: bug(populate): missing childKey causes no population

*bug*

Trying with 2.0.0 latest version.

When trying to pass a set of child keys to populate if some of the keys does not exist in the firebase route none of the keys are populated.

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via [codesandbox](https://codesandbox.io/) or similar.**

I have a /users/uid route in firebase which has only "experiences" key. If non existent childkeys are included in the populates list existing keys should still be populated. 

Populate works fine when I try to initiate with 

```js 
const populates = [
  { child: 'experiences', root: '/allExperiencesList' },
]

const fbWrappedComponent = firebaseConnect(
  ({AuthProfile}) => {
    if(AuthProfile){
      return ([{path: `/users/${AuthProfile ? AuthProfile.uid : ''}`, populates}])
    }
  }
)(TopBarContainer)
```
but fails if I add a key which does not exist and try as follows,

```js
const populates = [
  { child: 'experiences', root: '/allExperiencesList' },
  { child: 'education', root: '/allEducationList' },
]

const fbWrappedComponent = firebaseConnect(
  ({AuthProfile}) => {
    if(AuthProfile){
      return ([{path: `/users/${AuthProfile ? AuthProfile.uid : ''}`, populates}])
    }
  }
)(TopBarContainer)
```
Tried enabling logging to "true" and clearly see that populate is not even triggered in the second case.

As far as I know it did work in previous versions.

JeremyPlease:
### Description

This is a minor update to `populateChild` function in `helpers.js` to returns `null` instead of `child`. This works because the returned value is used in `reduce` function that combines values. 

Previously, already written populates values could be overwritten when `child` was returned.
A new failing test was written to reproduce the issue and is now passing. I've also confirmed this fix in a private repo with `npm link`.

### Check List

- [x] All tests passing

### Relevant Issues

Fixes #277.

### DIFF
--- a/tests/unit/helpers.spec.js
+++ b/tests/unit/helpers.spec.js
@@ -6,6 +6,7 @@ const exampleData = {
     projects: {
       CDF: {
         owner: 'ABC',
+        category: 'cat1',
         notes: {
           123: true
         },
@@ -15,7 +16,8 @@ const exampleData = {
         }
       },
       GHI: {
-        owner: 'ABC'
+        owner: 'ABC',
+        category: 'cat2'
       },
       OKF: {
         owner: 'asdfasdf',
@@ -29,9 +31,20 @@ const exampleData = {
       },
       QRS: {
         owner: 'ABC',
+        category: 'cat1',
         nested: {
           owner: 'ABC'
         },
+        notes: {
+          123: true
+        },
+        collaborators: {
+          ABC: true,
+          abc: true
+        }
+      },
+      TUV: {
+        owner: 'ABC',
         notes: {
           123: true
         },
@@ -44,6 +57,14 @@ const exampleData = {
     users: {
       ABC: {
         displayName: 'scott'
+      }
+    },
+    categories: {
+      cat1: {
+        displayName: 'magic'
+      },
+      cat2: {
+        displayName: 'animals'
       }
     },
     notes: {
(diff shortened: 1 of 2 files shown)


JeremyPlease:
Hey @prescottprue, it looks like Travis has an unrelated test failing for node 8. I'm unable to reproduce this failure locally using the same node version at Travis (8.7.0). Could you please trigger a rebuild in Travis to see if it resolves itself?

tests/unit/helpers.spec.js
```diff
@@ -214,6 +235,21 @@ describe('Helpers:', () => {
             .property('displayName', 'scott')
         })
 
+        it('populates value even when others are missing', () => {
```
prescottprue: Great work adding the unit test along with it!

tests/unit/helpers.spec.js
```diff
@@ -6,6 +6,7 @@ const exampleData = {
     projects: {
       CDF: {
         owner: 'ABC',
+        category: 'cat1',
```
prescottprue: Forgot this data was in the file directly. Might end up moving this out to its own file.

Great job figuring it out anyway :)

JeremyPlease:
Thanks, @prescottprue!

PR #299, opened 2017-10-20
feat(core): non-HOC version of firebaseConnect, for use with server side rendering

marekolszewski:
### Description

This PR adds a new `connect` helper to the props.firebase object that lets users connect their redux store to the firebase DB without requiring the firebaseConnect HOC. This is useful for populating your redux state ahead of rendering your React app, a typical pattern when performing server side rendering.

### Check List

- [X] All tests passing

### DIFF
--- a/src/createFirebaseInstance.js
+++ b/src/createFirebaseInstance.js
@@ -1,5 +1,6 @@
 import { isObject } from 'lodash'
 import { authActions, queryActions, storageActions } from './actions'
+import { getEventsFromInput, createCallable } from './utils'
 
 /**
  * Create a firebase instance that has helpers attached for dispatching actions
@@ -240,6 +241,27 @@ export const createFirebaseInstance = (firebase, configs, dispatch) => {
     storageActions.deleteFile(dispatch, firebase, { path, dbPath })
 
   /**
+   * @description Connect. Similar to the firebaseConnect Higher Order
+   * Component but presented as a function. Useful for populating your redux
+   * state without React, e.g., for service side rendering.
+   * @param {Array} watchArray - Array of objects or strings for paths to sync
+   * from Firebase. Can also be a function that returns the array. The function
+   * is passed the props object specified as the next parameter.
+   * @param {Object} props - The props object that you would like to pass to
+   * your watchArray generating function.
+   * @return {Promise}
+   */
+  const connect = (watchArray, props) => {
+    const inputAsFunc = createCallable(watchArray)
+    const prevData = inputAsFunc(props, firebase)
+    const firebaseEvents = getEventsFromInput(prevData)
+    const promises = firebaseEvents.map(event =>
+      queryActions.watchEvent(firebase, dispatch, event))
+
+    return Promise.all(promises)
+  }
+
+  /**
    * @description Watch event. **Note:** this method is used internally
    * so examples have not yet been created, and it may not work as expected.
    * @param {String} type - Type of watch event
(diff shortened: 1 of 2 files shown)


prescottprue:
Nice feature!

I actually started on some examples that include doing this by hand with `watchEvent`, but it is much nicer to have a helper like this.

One concern is the unmounting of the listeners added by `watchEvent`. We could handle this with notes in the docs, but it might be nice to think about an easy way to handle it.

We may have to come up with a nicer name for export though since `connect` is used by `react-redux`, which is commonly integrated with `react-redux-firebase`.

src/createFirebaseInstance.js
```diff
@@ -240,6 +241,27 @@ export const createFirebaseInstance = (firebase, configs, dispatch) => {
...
+   * is passed the props object specified as the next parameter.
+   * @param {Object} props - The props object that you would like to pass to
+   * your watchArray generating function.
+   * @return {Promise}
+   */
+  const connect = (watchArray, props) => {
+    const inputAsFunc = createCallable(watchArray)
+    const prevData = inputAsFunc(props, firebase)
+    const firebaseEvents = getEventsFromInput(prevData)
+    const promises = firebaseEvents.map(event =>
```
prescottprue: Kinda a nitpick: We should be able to use `watchEvents` and just pass the listener event settings array right into it instead of needing to call the `map` here.

src/createFirebaseInstance.js
```diff
@@ -407,7 +429,8 @@ export const createFirebaseInstance = (firebase, configs, dispatch) => {
     watchEvent,
     unWatchEvent,
     reloadAuth,
-    linkWithCredential
+    linkWithCredential,
+    connect
```
prescottprue: What about a different name for this export since `react-redux` uses `connect` and people commonly integrate with that.

src/createFirebaseInstance.js
```diff
@@ -240,6 +241,27 @@ export const createFirebaseInstance = (firebase, configs, dispatch) => {
     storageActions.deleteFile(dispatch, firebase, { path, dbPath })
 
   /**
+   * @description Connect. Similar to the firebaseConnect Higher Order
```
prescottprue: Great work with the comments here.

src/createFirebaseInstance.js
marekolszewski: That's what I had initially but watchEvents doesn't return a promise, which is important for SSR as you have to wait for the data to be fetched before rendering. I could update watchEvents if you want.

src/createFirebaseInstance.js
marekolszewski: Sure, happy to change it. How about firebaseConnectHelper or connectToEvents (to match watchEvents)?

prescottprue:
This has been renamed to `promiseEvents` to more closely match `watchEvent` since it is exposed on the firebase instance itself. It looked a little weird to have `firebase.firebaseWatch`

greaveselliott:
Hi @prescottprue,

I'm attempting to preload my redux state from firebase using ```store.firebase.promiseEvents()```.

Even though the promise resolves, the firebase data is never pulled and the request hangs, I've included a snapshot below of my store's state when the promise resolves.

```
{ 
requesting:
   { once: true,
     flamelink: true,
     'flamelink/environments/production/content/blog/en-US': true },
  requested:
   { once: false,
     flamelink: false,
     'flamelink/environments/production/content/blog/en-US': false },
  timestamps:
   { once: 1522711881049,
     flamelink: 1522711881049,
     'flamelink/environments/production/content/blog/en-US': 1522711881052 },
  data: {}
```
Appreciate your time.

prescottprue:
@greaveselliott A little hard to tell from what you are posting. Lets open a new issue with your exact use case with the listeners that are being set as well as any rules you have to make sure it is correctly replicated.

PR #305, opened 2017-10-24
Add getValueAt function for easy migration to v2 from dataToJS and pathToJS

fej-snikduj:
### Check List

- [ ] All tests passing

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -51,20 +51,16 @@ import { getDotStrPath } from './reducer'
  *   todos: getValueAt(firebase, 'data/todos/user1', defaultValue)
  * })
  */
-export const getValueAt = (data, path, notSetValue) => {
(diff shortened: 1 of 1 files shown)


src/helpers.js
```diff
@@ -16,6 +16,59 @@ import {
...
+ *   todos: getValueAt(firebase, 'data/todos/user1', defaultValue)
+ * })
+ */
+export const getValueAt = (firebase, path, notSetValue) => {
+  if (!firebase) {
+    return notSetValue
+  }
+
+  const dotPath = getDotStrPath(path)
+  const valueAtPath = get(firebase, dotPath)
```
prescottprue: We should also have the `notSetValue` be passed here so it is used as the defaultValue:

```js
return get(firebase, dotPath, notSetValue)
```

PR #309, opened 2017-10-26
Add EMPTY_AUTH_CHANGE case to Profile Reducer

fej-snikduj:
### Description
Adds EMPTY_AUTH_CHANGE case to profile reducer, so that firebase.profile is also updated to `{ isLoaded: true, isEmpty: true }` on empty auth changes.

### Check List

- [ ] All tests passing


prescottprue:
Nice work. I am also going to make `enableEmptyAuthChanges: true` be the default.

PR #313, opened 2017-10-29
Remove deep property for REMOVE action instead of setting it to undefined

fej-snikduj:
### Description
setWith sets the value to undefined, but the key remains an enumerable
property on the object.  This can cause problems when your UI iterates
over keys of an object and renders based on the key’s existence.  Using
unset, the key is removed.

### Check List

- [ ] All tests passing


prescottprue:
Great work, the `recursiveUnset` function is nice. Going to still make `dispatchRemoveAction: false` the default. That way it would require a listener on a given path in order to have data removed, unless this option is enabled. It will more closely match v1.

Issue #325, opened 2017-11-03
Routing Recipes - Some inconsistencies


jeandavid:
I have noticed some inconsistencies within `docs/routing/recipes.md` page.

* In the Advanced Section - Auth Required, for react-router v4 + redux-auth-wrapper v2, you import 
`import { browserHistory } from 'react-router'` but there is no `browserHistory` in react-router v4.

* In the Advanced Section - Redirect Authenticated, for react-router v4 + redux-auth-wrapper v2, you import `import { UserAuthWrapper } from 'redux-auth-wrapper'` but it should import `connectedRouterRedirect` for v2.

Also it would be great if the examples could be updated to show how to use react-router v4 + redux-auth-wrapper v2, not only on the redux-auth-wrapper higher order component, but also on the createStore.js and App.js file. 

Thanks!

NathHorrigan:
Hello! I would like to add to this discussion regarding an error I am having trying to implement this.
I have the following decorator (copied from docs) that doesn't have reference to firebase and so doesn't compile. Am I being stupid here?? I can't use `getFirebase` so I'm not sure what to do 😕 

```
// Decorator file
const Auth = connectedRouterRedirect({
  redirectPath: (state, ownProps) =>
  locationHelper.getRedirectQueryParam(ownProps) || '/login',
  allowRedirectBack: true,
  authenticatedSelector: ({ firebase }) => pathToJS(firebase, 'auth') !== null,
  authenticatingSelector: ({ firebase: { auth } }) =>
  pathToJS(firebase, 'isInitializing') === true ||
  pathToJS(firebase, 'auth') === undefined,
  AuthenticatingComponent: <div />,
  wrapperDisplayName: 'UserIsAuthenticated',
})
export default Auth
```

```
// How decorator is used
@firebaseConnect()
@Auth
export default class Home extends Component {
// etc...
```

The error I'm getting is: 
```
Failed to compile

./src/decorators/auth.js
  Line 51:  'firebase' is not defined  no-undef
  Line 52:  'firebase' is not defined  no-undef
```

prescottprue:
@NathHorrigan Looks like you are using the `v1.*.*` syntax and there is an issue in the `authenticatingSelector` (`({ firebase: { auth } }) ` should be, `({ firebase }) `).

prescottprue:
Routing recipes docs have been updated. The examples update will be coming at some point, but if you have time to capture it in a pull request, totally open to it.

Issue #320, opened 2017-11-01
bug(firebaseConnect): inconsistent arguments for the callback


optimistiks:
**Do you want to request a *feature* or report a *bug*?**  

A bug.

**What is the current behavior?**

Currently, on `componentWillMount`, first argument is `props`, second argument is `store`.
https://github.com/prescottprue/react-redux-firebase/blob/v2.0.0-beta.14/src/firebaseConnect.js#L47

But on `componentWillReceiveProps`, first argument is `props`, but the second argument is `firebase`.
https://github.com/prescottprue/react-redux-firebase/blob/v2.0.0-beta.14/src/firebaseConnect.js#L65

**What is the expected behavior?**

Callback function signature should be the same for both cases. `(props, store) => {}`

prescottprue:
You are correct, Great catch! Seems like the recently release that changed it didn't change the `componentWillRecieveProps`. Going to try to add a unit test to catch this in the future.

prescottprue:
The [v2.0.0-beta.15 release](https://github.com/prescottprue/react-redux-firebase/releases/tag/v2.0.0-beta.15) includes the fix for this.

Post if this or anything else doesn't work as expected. Thanks for reporting!

Issue #319, opened 2017-10-31
feat(auth): signInWithPhoneNumber


jacobfunch:
It would be much appreciated if react-redux-firebase supported Firebase signin with phone number (https://firebase.google.com/docs/auth/web/phone-auth).

Remember that it requires reCAPTCHA.

prescottprue:
I agree, I'll make an action creator.

@jacobfunch Do you plan to [use the invisible Captcha](https://firebase.google.com/docs/auth/web/phone-auth#use-invisible-recaptcha) attached to a button? Either way, that might be what I end up using for an example.

It should be able to work like so:

```js
// appVerifer is recaptcha-verifier
firebase.signInWithPhoneNumber(phoneNumber, appVerifier)
```


Good news: There may be a way to extend Firebase's functions with dispatch (still looking into it), so in the future, hopefully calling methods directly can work as expected.

jacobfunch:
@prescottprue perfect! And yes indeed, I plan on using the invisible reCAPTCHA – which I think most products will use as it is less intrusive.

prescottprue:
The first swing at it is on [the v2.0.0-beta.15 branch](https://github.com/prescottprue/react-redux-firebase/tree/v2.0.0-beta.15) (not yet published to NPM). Feel free to npm link and test it yourself, but I have yet to write tests for it or do much testing myself.

prescottprue:
`signInWithPhoneNumber` action creator added in [v2.0.0-beta.15](https://github.com/prescottprue/react-redux-firebase/releases/tag/v2.0.0-beta.15). It was interesting to integrate with the existing profile stuff, so let me know if things do not function as expected.

jacobfunch:
Thanks @prescottprue! When signing in after a user has been created it overwrites the entire user object even though there was custom data saved. Can one use the merge option?

prescottprue:
@jacobfunch Are you referring to the `auth` or `profile` object be overwritten? Is it the `SET_PROFILE` action that is overwriting it or another action type?

prescottprue:
@jacobfunch There is no merge option yet, but since #366 is asking for that, we can discuss the feature in that issue.

Thanks for your patience and clarifying your use case.

Issue #301, opened 2017-10-22
bug(profile) inconsistent profile isLoaded, isEmpty states


Modestas:
**Do you want to request a *feature* or report a *bug*?**  
Bug

**What is the current behavior?**
In one of the prop states during login process, I get such result:

**no auth:**
 isLoaded(auth) `true` isEmpty(auth) `true` isLoaded(profile) `true` isEmpty(profile) `true`
**after login pressed**
 isLoaded(auth) `false` isEmpty(auth) `true` isLoaded(profile) `false` isEmpty(profile) `true`
**[corrupted]** 
isLoaded(auth) `true` isEmpty(auth) `false` isLoaded(profile) `true` isEmpty(profile) `true`
**on next prop, it is auto fixed:**
isLoaded(auth) `true` isEmpty(auth) `false` isLoaded(profile) `true` isEmpty(profile) `false`

However this corrupted state means that user is authenticated but profile is not found (profile path existed in database even before login).

**What is the expected behavior?**
Corrupted state should not occur. In example, in my app I perform redirect if user is `auth` but does not have an entry in database.

**Which versions of dependencies, and which browser and OS are affected by this issue? Did this work in previous versions or setups?**
2.0.10

prescottprue:
@Modestas Going to investigate further, but from what you are reporting this seems like a bug. Thanks for reporting.

prescottprue:
@Modestas Upon further investigation it seems that this may be expected because you will first be logged in, then your profile listener will be applied, then finally the profile response will come back.

That means that `auth` will always come back before `profile`.

Could you look at which actions are being dispatched in redux devtools so we can confirm?

Modestas:
Action: `{ type: '@@reactReduxFirebase/LOGIN', auth: {...} }`
State: `auth: { isLoaded: true, isEmpty: false }`, `profile: { isLoaded: true, isEmpty: true }`

Action: `{ type: '@@reactReduxFirebase/SET_PROFILE', profile: {...} }`
State: `auth: { isLoaded: true, isEmpty: false }`, `profile: { isLoaded: true, isEmpty: false }`

Maybe this is fine, but in that `LOGIN` moment, state is kind of **invalid,** because it says that `profile` is indeed loaded but empty. In my opinion, login initialization should invalidate `profile`. Yes, it was loaded before login was initialized, but the moment login was started, it should change to `{ isLoaded: false }`.

[...]

PR #326, opened 2017-11-05
v2.0.0 beta.15

prescottprue:
### Description
* feat(auth): `UNLOAD_PROFILE` is dispatched when `login` action creator is called  - #301
* feat(auth): `signInWithPhoneNumber` action creator added - #319
* fix(firebaseConnect): function passed receives consistent arguments regardless of lifecycle - #320
* fix(query): `queryId` now includes `storeAs` to fix issue with listeners - #294
* fix(populate): ordered (array data) correctly populating - #239 - thanks @0x80 
* fix(reducer): property removed from profile is removed from state - thanks @fej-snikduj 
* fix(docs): inconsistencies in redux-auth-wrapper example - #325

### Check List

- [X] All tests passing

### Relevant Issues
* #239
* #294
* #301 
* #319
* #320
* #325

### DIFF
--- a/src/firebaseConnect.js
+++ b/src/firebaseConnect.js
@@ -62,7 +62,7 @@ export const createFirebaseConnect = (storeKey = 'store') =>
         componentWillReceiveProps (np) {
           const { firebase, dispatch } = this.store
           const inputAsFunc = createCallable(dataOrFn)
-          const data = inputAsFunc(np, firebase)
+          const data = inputAsFunc(np, this.store)
 
           // Handle a data parameter having changed
           if (!isEqual(data, this.prevData)) {
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ sudo: false
 language: node_js
 
 node_js:
-  - 6.11.3 # L.T.S
-  - 8 # Current
+  - 6.11.3 # Still Used
+  - 8 # L.T.S
+  - 9 # Current
 
 notifications:
   # Only send notifications when travis status changes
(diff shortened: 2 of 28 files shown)


src/firebaseConnect.js
```diff
@@ -62,7 +62,7 @@ export const createFirebaseConnect = (storeKey = 'store') =>
         componentWillReceiveProps (np) {
           const { firebase, dispatch } = this.store
           const inputAsFunc = createCallable(dataOrFn)
-          const data = inputAsFunc(np, firebase)
+          const data = inputAsFunc(np, this.store)
```
fej-snikduj: Whats this change about?

prescottprue: To provide `store` as the second argument to functions passed to `firebaseConnect`. That way, `store.getState` can be used without needing to use a preceding `connect` call.

prescottprue: It is mentioned in [the migration guide](http://docs.react-redux-firebase.com/history/v2.0.0/docs/v2-migration-guide.html#stateBasedQueries)

fej-snikduj: Ok - seems like it would be breaking if you were expecting firebase as the second argument.  We don't use a second argument anywhere, so it doesn't matter to us

prescottprue: Yup, totally agree. It will be mentioned as a breaking change in the release. It is still available as `store.firebase`, but I was also thinking of making a config option that would allow users to keep things as they were.

PR #328, opened 2017-11-05
feat(populate): populate working for ordered data (arrays)

0x80:


0x80:
I replaced some use of Lodash map & reduce with native ES ones. I think it makes the code more readable.

prescottprue:
Nice work! Reviewing and testing today. Id like to get tests for this in before taking v2.0.0 out of pre-release, but if all is well we can merge this without the tests and worry about it later.

0x80:
I'm afraid I can't allocate more time for writing tests in the coming days. I already spent half a day of my client's time making the PR, and we're behind on work as it is :)

PR #331, opened 2017-11-08
Fix promisesForPopulate to fetch data in all cases

fej-snikduj:
### Description
Fixes population to work for cases when a non-existant child key is included alongside existing child keys.

### Check List

- [X] All tests passing


src/helpers.js
```diff
@@ -280,7 +280,7 @@ export const populate = (state, path, populates, notSetValue) => {
 
   if (!isArray(data)) {
     // check each populate child parameter for existence
-    const dataHasPopulateChilds = every(populatesForData, p =>
+    const dataHasPopulateChilds = some(populatesForData, p =>
```
prescottprue: Yes! This is exactly what I thought it was going to be. Nice work!

yarn.lock
```diff
@@ -0,0 +1,8254 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
prescottprue: Can we remove this? v2.0.0 has yarn removed in favor of `package-lock.json` (node >8) since apparently it is faster or at least equivalent. Node 9 has been added to travis to confirm that builds work and tests pass correctly with that version too.

tests/unit/utils/populate.spec.js
```diff
@@ -80,6 +80,12 @@ describe('Utils: Populate', () => {
       expect(res).to.have.deep.property(`users.${uid}`)
     })
 
+    it('populates all existing children even if one populates child does not exist', async () => {
+      populates = [{child: 'collaborators', root: 'users'}, {child: 'nonExistantKey', root: 'users'}]
```
prescottprue: GREAT WORK! So easy to read and understand

src/enhancer.js
```diff
@@ -50,7 +50,7 @@ let firebaseInstance
  * @param {Array|String} config.profileParamsToPopulate - Parameters within
  * profile object to populate. As of `v2.0.0` data is only loaded for population, not actually automatically populated
  * (allows access to both unpopulated and populated profile data).
- * @param {Boolean} config.autoPopulateProfile - **NOTE**: Not yet enabled for v2.0.0. Whether or not to
+ * @param {Boolean} config.autoPopulateProfile - **NOTE**: Now enabled for v2.0.0. Whether or not to
```
prescottprue: Another great catch! Good stuff.

Issue #310, opened 2017-10-27
bug(populate): undefined child causing expected behavior


aQustic:
**What is the current behavior?**

Every object of an array which does not have the child element defined in population rule gets transformed into an object full of arrays.

```
const populates = [
  { root: 'messages', child: 'messages' },
];
```

```
@firebaseConnect(
  ({ params }) =>
    [{ path: `groups/${params.slug}`, populates }]
)
@connect(
  ({ firebase }, { params }) => ({
    group: populate(firebase, `groups/${params.slug}`, populates),
  })
)
```

This produce the expected population
from
```
group
 - createdAt: 1509...
 - fromID: "XZ2ovIPB..."
 - messages: {...}
  -- KxSP... : true
```
to
```
group
 - createdAt: 1509...
 - fromID: "XZ2ovIPB..."
 - messages: {...}
  -- KxSP... : {...}
   --- fromID: "...."
   --- text: "...."
```

but if a group does not have a child array of messages it produces strange arrays within the group object for every attribute it has:
```
group
 - createdAt: {...}
  -- Empty Array
 - fromID: {...}
  -- 0: "X"
  -- 1: "Z"
  -- 2: "2"
  -- 3: "o"
  -- 4: "v"
 - name: {...}
  -- 0: "T"
  -- 1: "e"
  -- 2: "s"
  -- 3: "t"
```

In addition to that this arrays are returned from strings while integers are returning empty arrays.

**What is the expected behavior?**

Original state of object will not be modified. Only the child array of keys should be populated, if it exist.

prescottprue:
@aQustic This is for sure not expected. I am going to try to replicate myself to confirm and see what could be causing it.

prescottprue:
@aQustic Have you happened to confirm this on v2.0.0-beta.14 as well?

aQustic:
@prescottprue i've updated to beta.14 but the same error appears.

The console says
```
Uncaught (in promise) Error: Objects are not valid as a React child (found: object with keys {0, 1, 2}).
```
and the props of the object contain the same arrays i've experienced with beta.13 if populated array don't exist 
```
group
 - createdAt: {...}
  -- Empty Array
 - fromID: {...}
  -- 0: "X"
  -- 1: "Z"
  -- 2: "2"
  -- 3: "o"
  -- 4: "v"
 - name: {...}
  -- 0: "T"
  -- 1: "e"
  -- 2: "s"
  -- 3: "t"
```

[...]

Issue #302, opened 2017-10-22
bug(docs): queries including startAt, endAt, equalTo & limits with orderByValue are order dependent


AVVS:
**What is the current behavior?**

```
        case 'startAt':
          let startAtParam = !doNotParse ? tryParseToNumber(param[1]) : param[1]
          startAtParam = startAtParam === 'null' ? null : startAtParam
          query = param.length === 3
            ? query.startAt(startAtParam, param[2])
            : query.startAt(startAtParam)
          break
        case 'endAt':
          let endAtParam = !doNotParse ? tryParseToNumber(param[1]) : param[1]
          endAtParam = endAtParam === 'null' ? null : endAtParam
          query = param.length === 3
            ? query.endAt(endAtParam, param[2])
            : query.endAt(endAtParam)
          break
```

When query is:

```js
queryParams: ['orderByValue', `limitToFirst=${limitToFirst}`, `startAt=${startAt}`],
```

`limitToFirst` and `startAt` would be treated as string failing the query. It would work if we do

```js
queryParams: [`limitToFirst=${limitToFirst}`, `startAt=${startAt}`, 'orderByValue'],
```

**What is the expected behavior?**

`orderByValue` does not turn on `doNotParse` on

**Which versions of dependencies, and which browser and OS are affected by this issue? Did this work in previous versions or setups?**

Branch 2.x.x

prescottprue:
Good find @AVVS, the order shouldn't matter, so this is for sure a bug. I'll look into fixing it, but also open to pull requests if anyone gets to it.

It may be irrelevant, and should definitely be documented more, but did you try out the `noParse` param?

prescottprue:
Query docs have been updated with a note about how order is important, as well as how to use `notParsed`.

In v2.0.0-beta.17 I have planned on adding a `parsed` query parameter. That way, you will be able to pass your params out of order, and enable/disable automatic parsing as needed.

Let me know if this doesn't address your concerns, and we can reopen. Thanks for reporting!

PR #334, opened 2017-11-10
v2.0.0-beta.17

prescottprue:
### Description
* fix(populate): undefined populate child responds unmodified as expected - #310
* fix(auth): `UNLOAD_PROFILE` action type no longer dispatched (did not fix issue) - #301
* feat(populate): support for populating ordered data with multiple populates where one populate child missing
* feat(query): parsed query param added to allow disabling of automatic parsing - #302
* feat(config): `preserveOnLogin` config option added to allow preserving of pieces of state affected by `LOGIN` action (called by `login` action creator)
* feat(tests): test added to check profile state consistency on `LOGIN` dispatch - #301
* feat(tests): tests added for `getVal` helper
* feat(tests): Unit tests added for populating ordered data

### Check List

- [X] All tests passing

### Relevant Issues
* #301
* #302
* #310


src/reducer.js
```diff
@@ -250,7 +249,11 @@ export const authReducer = (state = { isLoaded: false, isEmpty: true }, action)
         }
       }
       const auth = action.auth.toJSON ? action.auth.toJSON() : action.auth
-      return { ...auth, isEmpty: false, isLoaded: true }
+      // Support keeping data
+      if (action.preserve && action.preserve.auth) {
+        return pick({ ...state, ...auth }, action.preserve.auth) // pick returns a new object
+      }
+      return { ...state, ...auth, isEmpty: false, isLoaded: true }
```
fej-snikduj: before we were returning `{ ...auth, isEmpty: false, isLoaded: true }` and now we are returning `{ ...state, ...auth, isEmpty: false, isLoaded: true }`.  Any reason we added `...state`?

prescottprue: Great question! The goal was to preserve anything in state that was there, as well as spread the new parameters onto it. I think that this doesn't really offer that much advantage, and could be unclear to some, so I am going to change it back.

I think it still makes sense to pass `{ ...state, ...auth }` to `pick` though since some want to be able to preserve `isLoaded` state and potentially other things that were already in state.

PR #371, opened 2018-01-10
fixed missing babel presets needed in devDependencies

hcwiley:
### Description
I just checked out this project and was not able to build the material example. realized it was just missing the babel presets for stage-0 and es2015. i guess other people have those globally installed and isn't an issue for others?

### Check List

- [x] All tests passing

### DIFF
--- a/examples/complete/material/package-lock.json
+++ b/examples/complete/material/package-lock.json
@@ -463,6 +463,17 @@
         "trim-right": "1.0.1"
       }
     },
+    "babel-helper-bindify-decorators": {
+      "version": "6.24.1",
+      "resolved": "https://registry.npmjs.org/babel-helper-bindify-decorators/-/babel-helper-bindify-decorators-6.24.1.tgz",
+      "integrity": "sha1-FMGeXxQte0fxmlJDHlKxzLxAozA=",
+      "dev": true,
+      "requires": {
+        "babel-runtime": "6.26.0",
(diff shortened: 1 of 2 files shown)


prescottprue:
@hcwiley The material example actually does not use those presets anymore [it now uses `babel-preset-env`](https://github.com/prescottprue/react-redux-firebase/blob/master/examples/complete/material/package.json#L57). It looks like the `.babelrc` is still there when it should have been removed (babel settings are in build folder).

As [noted in the comment at the top of the `.babelrc` file](https://github.com/prescottprue/react-redux-firebase/blob/master/examples/complete/material/.babelrc#L1-L4), the `.babelrc` file being there is just to prevent an issue some users had with a global `.babelrc` being used.

The best course of action is probably just to update the `.babelrc` to no include anything (just empty config). That way hopefully it is less confusing (understandable that seeing those presets in there makes one think they are being used).

prescottprue:
Just pushed up changes which include removing the `.babelrc` file and including `babelrc: false` to prevent global babelrc files [as described in this babel discuss post](http://discuss.babeljs.io/t/solved-how-do-we-make-babel-ignore-babelrc-when-used-in-webpack/147/4).

Thanks for reporting!

hcwiley:
ok cool! thanks. I'm not as up to speed with react stuff as I used to be, so not sure the right approach, just knew it wasn't working as intended!

Issue #372, opened 2018-01-11
How to query user specific data within firebaseConnect (incl. population)


ralcon:
Hi,

i'm working with react-native and have the problem to load only user specific data with `firebaseConnect`. Imagine there are thousends of `todos`, `users` and `doers` and don't want to reference all elements of them.

First of all, a shortcut of my data structure in the real time database:
```
- todos
  |- cleanRoom
  |    |- createdBy: alica
  |    |- title: "Clean the room up"
  |    |- doerList
  |        |- roomCleanerMichael
  |        |- roomCleanerAlica
  |- playBall
       |- createdBy: michael
       |- title: "Play the ball"
       |- doerList
           |- ballPlayerAlica

- users
  |- michael
  |    |- name: Michael
  |    |- age: 22
  |- alica
       |- name: Alica
       |- age: 27

- doers
  |- roomCleanerMichael
  |    |- user: michael
  |    |- status: done
  |- roomCleanerAlica
  |    |- user: alica
  |    |- status: done
  |- ballPlayerAlica
       |- user: alica
       |- status: done
``` 
I have a view/screen in RN, where all informations should be displayed. For now, i have...
```
const populatesTodoCreator = [{child: 'createdBy', root: 'users'}]

const mapStateToProps = (state, props) => {
  return {
    todos: populate(state.firebase, 'todos', populatesTodoCreator),
    uid: state.firebase.auth.uid
  }
}

export default compose(
  firebaseConnect((props, store) => {
      return [
        {path: 'todos', populatesTodoCreator},
      ]
    }
  ),
  connect(mapStateToProps, mapDispatchToProps)
)(TodoOverviewScreen)
```
When i log in the application with `firebase.auth.uid = michael`, i only want to reference the list of `todos`, where `michael` participates in as `doer` and populate all `doers` in `doersList` (including the `user` in a specific `doer`), and that all in only one single `compose` method.

So the redux state should look like:
```
- todos
  |- cleanRoom
      |- createdBy
      |    |- name: Alica
      |    |- age: 27
      |- title: "Clean the room up"
      |- doerList
           |- roomCleanerMichael
           |     |- user
           |     |    |- name: Michael
           |     |    |- age: 22
           |     |- status: done
           |- roomCleanerAlica
                 |- user
                 |    |- name: Alica
                 |    |- age: 27
                 |- status: done
```
Is this task possible? If yes, how do i get this done? Any structural suggestions?

[...]

PR #398, opened 2018-02-04
Update documentation of property population.

ralcon:
(as described in https://github.com/prescottprue/react-redux-firebase/issues/372#issuecomment-356801514)

### Relevant Issues
#372

### DIFF
--- a/docs/populate.md
+++ b/docs/populate.md
@@ -10,18 +10,18 @@ import { compose } from 'redux'
 import { connect } from 'react-redux'
 import { firebaseConnect, populate } from 'react-redux-firebase'
 
-const populates = [
+const populatesOwnerWithUser = [
   { child: 'owner', root: 'users' } // replace owner with user object
 ]
 
 const enhance = compose(
   firebaseConnect([
     // passing populates parameter also creates all necessary child queries
-    { path: 'todos', populates }
+    { path: 'todos', populates: populatesOwnerWithUser }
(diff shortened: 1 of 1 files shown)


prescottprue:
Not sure I understand what is being changed here or its effect?

Looked at the comment, and it seems I just pointed out how to use multiple populates. Then when looking at the diff of the commit I didn't see any multiple populates, just renaming of variables in a way that shouldn't change anything.

docs/populate.md
```diff
@@ -10,18 +10,18 @@ import { compose } from 'redux'
...
-const populates = [
+const populatesOwnerWithUser = [
   { child: 'owner', root: 'users' } // replace owner with user object
 ]
 
 const enhance = compose(
   firebaseConnect([
     // passing populates parameter also creates all necessary child queries
-    { path: 'todos', populates }
+    { path: 'todos', populates: populatesOwnerWithUser }
```
prescottprue: What does this renaming do here? Is it intended to more clearly indicate the name `populates`?

ralcon:
My fault. Something like JS newbe. Now understanding passing properties.
Passing `{ populates }` is the same like `{ populates: populates }`.

prescottprue:
@ralcon Glad you reached out anyway! The examples/docs use some short hand like that to keep things clean, but it can make some other things easy to miss (hard balance to strike).

Issue #401, opened 2018-02-06
[bug] LOGIN is called after SET_PROFILE (Logout/Login) issue


evgeni-tsn:
Just submitted a pull request for the issue below -> #402 

I have a react app with **login / logout** functionality using this library.
I noticed a strange bug which I tracked down and I have a solution for it.

If you are logged in and you logout then relogin again without refreshing the web page, the `firebase.profile` object stays empty.

That is happening only if we go from: 
**logged in -> logged out -> logged in again**.
![image](https://user-images.githubusercontent.com/7677924/35882243-280cd3ee-0b8c-11e8-9e61-d8a032bbc756.png)

After I checked the two actions that are dispatched (**LOGIN** and **SET_PROFILE**) it looks that the SET_PROFILE is called first and sets the profile object right and after that the **LOGIN** is being dispatched and it wipes the profile data. Here are the screenshots.
The objects looks good after SET_PROFILE
![image](https://user-images.githubusercontent.com/7677924/35882378-acda4e76-0b8c-11e8-9739-99c60fac5b8b.png)
And then it is cleared by LOGIN
![image](https://user-images.githubusercontent.com/7677924/35882413-ca532ffe-0b8c-11e8-9802-71a7d7c2e159.png)

Although this is not happening if open the web app in logged out state and just login.
After some debugging I found where the issue is. 
In **actions/auth.js** in **handleAuthStateChange** currently the library is dispatching **LOGIN** after the **watchUserProfile()**.

If we replace the order of that two all goes well and the issue is fixed.

![image](https://user-images.githubusercontent.com/7677924/35883039-a453d464-0b8e-11e8-96d4-6e7d95d3ca23.png)

I can submit a PR for that if you agree that there is an issue.

Windows 10
Versions:
    "react": "16.2.0",
    "react-redux-firebase": "2.0.3",

PR -> #402

prescottprue:
Included in [`v2.0.4` release](https://github.com/prescottprue/react-redux-firebase/releases/tag/v2.0.4). Thanks for reporting and providing the PR!

PR #402, opened 2018-02-06
fix(login): login-logout-login flow always correctly updates firebase.profile - #401 - @evgeni-tsn

evgeni-tsn:
### Description
Here is the full explanation to the issue: #401 

### Check List

- [x] All tests passing

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -345,13 +345,13 @@ const handleAuthStateChange = (dispatch, firebase, authData) => {
       setupPresence(dispatch, firebase)
     }
 
-    watchUserProfile(dispatch, firebase)
-
     dispatch({
       type: actionTypes.LOGIN,
       auth: authData,
       preserve: config.preserveOnLogin
     })
+
+    watchUserProfile(dispatch, firebase)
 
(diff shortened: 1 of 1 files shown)


src/actions/auth.js
```diff
@@ -345,14 +345,14 @@ const handleAuthStateChange = (dispatch, firebase, authData) => {
       setupPresence(dispatch, firebase)
     }
 
-    watchUserProfile(dispatch, firebase)
-
```
evgeni-tsn: I've just change the order of these two function. The result is that the LOGIN action will be called before SET_PROFILE as it should be.

prescottprue:
Great work! Glad to see both the issue and the PR to fix it.

Going to look into it a bit more to confirm that this shouldn't be placed in v2.1.0 and/or have a setting to switch it back to how it was since some may have depended on how things were.

prescottprue:
Confirmed that there are no cases (at least covered in documentation as expected usage patterns) that are negatively effected by this change.

It seems like it is even overkill to have a config to switch it back since it is not expected behavior.

Thanks again!

Issue #349, opened 2017-12-01
example(populate): populate by object's keys


MikeKoval:
Hi!

I have a profile object:
```
    {
      "friends" : {
        "-Kp-ky6hOPq-nImOpVGk" : "gsTflZjxoYcdQJUrS6nw6uxLFa93",
        "-KyzHhAr6QgdAcqnE5VF" : "frByC4g8cJTxcCNjfL2TWUMj9aR2"
      },
      "matches" : {
        "2LA4QQHfMQhP6Gdk8iaEFvlN1Du2" : 14,
        "YxqFK8M2nJOKt0Wz17Fj8sTNY1u1" : 18,
      },
      "uid" : "GauAnEhTaNOepGJbEmlxTW7Ex7G3"
    }
```

To populate list of friends I just need to add:
```
@connect(
  (state) => ({
    profile: populate( state.firebase, 'profile',   [ { child: 'friends', root: 'users' }])
  })
)
```

How can I populate matches by keys instead of values or maybe by subfield? 
I just don't want to get the full list of users from db, I think there must be a better way.

kolj:
Also interested how to solve such kind of issues.

prescottprue:
Great to know there is interest in this.

If the values are `true` (i.e a `$key: true` list) it will populate automatically. If the values are not `true` (like in the example you provided) - you can provide a function that calculates you population settings (instead of just an array). I'll work on an example to show what I mean.

MikeKoval:
@prescottprue Thank you!

Do you mean https://github.com/prescottprue/react-redux-firebase/blob/master/src/utils/populate.js#L108 , https://github.com/prescottprue/react-redux-firebase/blob/master/src/helpers.js#L251 and https://github.com/prescottprue/react-redux-firebase/blob/master/src/helpers.js#L264 , am I right?

It'd be great if you could provide an example :slightly_smiling_face:

maxvashchuk:
I would also like to see an example

MikeKoval:
@prescottprue Could you please help me, how I can provide my own function calculating my population settings (instead of just an array) ?

dibondi:
I am working with firebase and would like to know a better way as well.

ctwhome:
Hi, I found myself doing the next, although it works, I am not sure if that is the way to go:
```
@connect(
  // mapStateToProps (state)
  state => {
    if (state.firebase.data.operations) {
      return {
        uid: state.firebase.auth.uid,
        operations:
[...]

PR #411, opened 2018-02-14
populate method: use populateBykey = true to find items by key

meksikann:
### Description
Hi. creating PR for issue described below.
I tried to solve it using callback as you wrote,but in vain.
Added on property check if true. It doesn't brake existing methods.
Tested on my app. Works fine!!

### Check List

### Relevant Issues
https://github.com/prescottprue/react-redux-firebase/issues/349

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -210,7 +210,7 @@ const buildChildList = (state, list, p) =>
         : get(state.data, pathString)
     }
     // Populate child does not exist
-      return val === true || p.populateByKey ? val : getKey;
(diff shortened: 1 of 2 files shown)


prescottprue:
Nice work! Going to test this out today and merge it in.

I might write a test for it before its published, but the feature itself looks good to me. Its nice to see it under another parameter to prevent breaking, but I'm wondering if this should happen automatically in some cases?

meksikann:
thanks :)
it would be great idea)

PR #435, opened 2018-03-25
fix(storage): fix a typo related to UploadTaskSnapshot within uploadFile

dpehrson:
Previously the typo `uploadTaskShaphot` was pervasive throughout the
variables names and return values of firebase storage upload tasks,
including in the objects returned in promise resolutions to the
end-user. This commit preserves the typos in the public interface while
also providing the correct `uploadTaskSnapshot` key in those objects.

### Description

I noticed that when handling the promise resulting from `firebase.uploadFile()` the object being returned was incorrectly providing a key named `uploadTaskSnaphot` instead of `uploadTaskSnapshot` after wondering why I kept destructuring out an undefined value.

### Check List

- [x] All tests passing

**I could not find any existing tests related to testing this return value, so I did not add any**


prescottprue:
Great find! Also glad to see the preservation of legacy.

dpehrson:
🍾 woo!

It may be worth making a note to remove the typo properties from the public interface in the next major version release.

Thanks for the quick PR accept! This library has been a breeze to use so far.

PR #442, opened 2018-03-31
Make Example Props match Proptypes

connormckelvey:
### Description

In the current example, the `Uploader` component destructures `deleteFile` directly off of `props`, but the `Uploader.propTypes` object shows that, `deleteFile` is a function on `firebase`.

### Check List

- [x] All tests passing


prescottprue:
Great catch, thanks for the PR.

PR #451, opened 2018-04-13
fix(watch): issue with multiple watchers on same path

Tapped:
When a path was unwatched, which had more than one watcher, all the watchers were unwatched. The issue was caused by an early out return in `actions/query.js:watchEvent`, instead of relying on the refcounting system in `utils/query.js`.

### Check List

- [x] All tests passing

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+singleQuote: true
+semi: false
+trailingComma: 'none'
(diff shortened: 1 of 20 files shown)


prescottprue:
Thanks for the contribution! Since this may be a breaking change for some we will have to look at which version it should be released in.

Tapped:
I'm not to decide, but I guess it make sense that it's part of v2.1, since as you said it's a breaking change. However, at the same time it's fixing a fundamental feature, and users of a library should expect that there are breaking changes when a bug is fixed, that's how bugs works. 

Also, it's not clear for me how existing users are depending on the old behaviour. In other words, I don't think it will cause any major issues for existing users.

santiago-elustondo:
yes, this worked for me too

Issue #467, opened 2018-05-15
Deprecated firebase.User.prototype.signInWithEmailAndPassword


jaschaio:
Getting this console warning when logging in with username and password:
```
Deprecated firebase.User.prototype.signInWithEmailAndPassword in favor of firebase.User.prototype.signInAndRetrieveDataWithEmailAndPassword.
```

Using `react-redux-firebase:2.1.1` together with `react-native-firebase:4.1.0` – not entirely sure which of the two is causing this warning but I suspect it's `react-redux-firebase` as `react-native-firebase` should only be a native wrapper of the normal `firebase` sdk API methods.

prescottprue:
Reproduced with new Firebase version.

Weirdly, when putting this simple logic in place, I got a warning that is telling me the exact opposite?

![image](https://user-images.githubusercontent.com/2992224/40360055-a18c904e-5d79-11e8-8c86-a77aeb081681.png)

Going to investigate more before making a change. Any input is welcome.

prescottprue:
@jaschaio Just to confirm which method did you call to sign in?

prescottprue:
[`v2.1.5`](https://github.com/prescottprue/react-redux-firebase/releases/tag/v2.1.5) should fix the issue for you thanks to a PR from @kadikraman. Reach out if you are still seeing the warning, thanks for reporting.

dustingraves:
I am also seeing `firebase.auth.Auth.prototype.signInAndRetrieveDataWithEmailAndPassword is deprecated. Please use firebase.auth.Auth.prototype.signInWithEmailAndPassword instead.`

```javascript
      emailLogin: ({ firebase, router, showError }) => creds =>
      firebase.login(creds).catch(err => showError(err.message)),
```

fredrik-sogaard:
I'm seeing this as well. Oddly only sometimes. Can't really figure out when this is happening.

fredrik-sogaard:
"firebase": "^5.0.4",
 "react-redux-firebase": "^2.1.5",

Now I seem to get it every time. Screenshot: https://cl.ly/0q2U0Z07012D

fredrik-sogaard:
I can try updating Firebase and react-redux-firebase to the latest release and see if that helps.

fredrik-sogaard:
It's the same with Firebase 5.2.0 and react-redux-firebase 2.1.6.

Roustalski:
[...]

PR #484, opened 2018-06-06
feat(auth): use new firebase login methods

kadikraman:
### Description

As of [react-native-firebase v3.2.3](https://github.com/invertase/react-native-firebase/releases/tag/v3.2.3), `signInWithCustomToken` is deprecated in favour of `signInAndRetrieveDataWithCustomToken`.

screen shot 2018-06-06 at 13 50 59

Found a similar issue [here](https://github.com/prescottprue/react-redux-firebase/issues/467) and implemented a fallback for users in the older version of firebase as suggested.

### Check List
- [ ] All tests passing

Fixes https://github.com/prescottprue/react-redux-firebase/issues/467

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -443,35 +443,31 @@ export const login = (dispatch, firebase, credentials) => {
 
   const { method, params } = getLoginMethodAndParams(firebase, credentials)
 
-  const firebaseAuth = firebase.auth()
-
-  let firebaseMethod = firebaseAuth[method]
-
-  // support older versions of firebase that still use signInWithCustomToken & signInWithEmailAndPassword
-  switch (method) {
-    case 'signInAndRetrieveDataWithCustomToken':
-      firebaseMethod = firebaseAuth.signInWithCustomToken
-      break
-    case 'signInAndRetrieveDataWithEmailAndPassword':
-      firebaseMethod = firebaseAuth.signInWithEmailAndPassword
-      break
-    default:
-      break
-  }
-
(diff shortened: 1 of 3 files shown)


src/actions/auth.js
```diff
@@ -443,21 +443,35 @@ export const login = (dispatch, firebase, credentials) => {
...
 
-  return firebase
-    .auth()
-    [method](...params)
+  const firebaseAuth = firebase.auth()
+
+  let firebaseMethod = firebaseAuth[method]
+
+  // support older versions of firebase that still use signInWithCustomToken & signInWithEmailAndPassword
+  switch (method) {
```
prescottprue: Is it possible for us to fit this logic in the `getLoginMethodAndParams` function?

test/utils.js
```diff
@@ -129,7 +129,7 @@ export const fakeFirebase = {
               email: 'test@test.com',
               providerData: [{}]
             }),
-    signInWithCustomToken: () => {
+    signInAndRetrieveDataWithCustomToken: () => {
```
prescottprue: Nice work updating the tests too!

kadikraman:
@prescottprue thanks for the feedback. Updated to move logic to `getLoginMethodAndParams` which incidentally also fixed the tests that were failing before.

I think this definitely needs a test, but I'm not sure how to test for different versions of firebase. Any ideas?

prescottprue:
@kadikraman Thanks for making the changes, and nice work!. I can look into good ways to test the different versions.

PR #507, opened 2018-07-18
fix(auth): payload of AUTH_UPDATE_SUCCESS action dispatch fixed

stefanhar:
### Description
Auth object is being reset on successful call to updateAuth. This forces users to assume logged out state.

```auth``` should be passed instead of ```payload``` on action type AUTH_UPDATE_SUCCESS

as seen on line 192 in src/reducers.js

### Check List

- [x] All tests passing


prescottprue:
I'll look into getting this into a version for release, thanks for posting the fix!

PR #514, opened 2018-07-31
Optimize firestoreConnect unset / set listeners

demoran23:
### Description
When your queries change (as may happen when the queries themselves are based on state), instead of killing all of your listeners and then resubscribing everything, only deal with stuff that needs to be added and needs to be removed.

This is particularly important when dealing firestore's `composite` slice, since the data in there is in sync with your queries rather than having out of sync legacy data.

### Check List
- [Y] All tests passing
- [N/A] Docs updated with any changes or examples if applicable
- [N] Added tests to ensure new feature(s) work properly

### Tests (spies)
I don't quite know how to use `chai`'s spy plugin to know whether `store.firestore.setListeners` is getting called, and with what.


prescottprue:
@demoran23 The `setListeners` exported from `redux-firestore` that you are describing is before [it is bound with dispatch when creating the instance](https://github.com/prescottprue/redux-firestore/blob/master/src/createFirestoreInstance.js#L33-L38).

This is done because `dispatch` is needed to dispatch actions. When creating your store you use the `reduxFirestore` and `reactReduxFirebase` store enhancers to "enhance" the store with `store.firestore` and `store.firebase` - those are the "instances" that are Firebase methods wrapped with the correct action dispatches. From there, store is passed into context by setting up `Provider` from react-redux. `withFirebase`, `firebaseConnect`, `withFirestore`, and `firestoreConnect` grab store from context and use it to access either `store.firebase` or `store.firestore`.

If you are interested, [here is the actual logic that passes dispatch to `createFirestoreInstance`](https://github.com/prescottprue/redux-firestore/blob/master/src/enhancer.js#L49-L64).

Love talking about this kind of stuff, so if you reach out [over gitter](https://gitter.im/redux-firebase/Lobby) we can have a quicker back and forth 😄

Issue #488, opened 2018-06-15
bug(storage): issue uploadFile when using meta information


reimertz:
It seems we are making an assumption here that FireStore add and Database push resulting in the same returned value. But it seems Firestore returns a snapshot where's database push only returns an error if something went wrong. 

https://github.com/prescottprue/react-redux-firebase/blob/51ce8dbd076002550208be2f03ac1bb7de630b66/src/utils/storage.js#L140

I assume this also relates to the issues seen in #487.

prescottprue:
Good find. Open to a PR if you get a chance, otherwise I will try to get a fix out this weekend.

reimertz:
Cool!

I'm trying to figure out how this code works. It seems we're doing `uploadResultFromSnap` after we've pushed `fileData` using 

```
...

var metaSetPromise = function metaSetPromise(fileData) {
    return useFirestoreForStorageMeta ? firebase.firestore().collection(dbPath).add(fileData) : firebase.database().ref(dbPath).push(fileData);
  };

return metaSetPromise(fileData).then(resultFromSnap);
```

So I don't understand the reasoning why we want to `.then(resultFromSnap)`. 

Is it because we want to be able to dispatch `FILE_UPLOAD_COMPLETE`? If so, it seems nothing is dependent on that action anyway?

prescottprue:
Yeah, glad you are finding the same thing. As mentioned in #487, that seems to be happening out of order and will need to be fixed.

jpierce42:
Thanks for all the work on this library - just started using it a couple weeks ago and it's amazing.  

I ran into the same problem described in #480 when trying to upload files, dug around a bit and ended up installing from the v3.0.0-alpha branch to get the latest code [adding `downloadURL` to the file metadata](https://github.com/prescottprue/react-redux-firebase/commit/c42936d2d995a0535094f265c1abc4916410d18f) .  I ran into an async-related error:

"ReferenceError: regeneratorRuntime is not defined" 

Seems like this can be fixed by installing babel/polyfill:  [https://github.com/babel/babel/issues/5085](https://github.com/babel/babel/issues/5085)

I installed that, and also updated the syntax of the arguments passed to `fileMetadataFactory`.  It looks like `downloadURL` is not part of the default `uploadRes` object, but comes in separately?

```
fileMetadataFactory: (uploadRes, firebase, metadata, downloadURL) => {
  const { metadata: { name, fullPath } } = uploadRes
  return {
    name,
[...]

Issue #487, opened 2018-06-12
feat(storage): make downloadURL available for file metadata (firestore/firebase v5)


nasimoyz:
**What is the current behavior?**
Thanks for all your amazing work on this library and for quickly deploying the fix for file upload (for firebase v5)!

I was wondering about the best way to write the `downloadURL` to the File Metadata when working with the firebase v5 updates?

Currently the File Metadata is written before the downloadURL is retrieved (with the downloadURL retrieval occurring in `uploadResultFromSnap`).
I can write it to the File Metadata after the `firebase.uploadFile` promise resolves, only issue is then writing File Metadata will always incur a second network call.

On dependencies:
```
"firebase": "^5.0.4",
"react-redux-firebase": "^2.1.5",
```

prescottprue:
@nasimoyz Yup! Glad the issues were reported so quickly!

As for getting `downloadURL` with v5: [this logic, which is new as of `v2.1.5`, gets the `downloadURL`](https://github.com/prescottprue/react-redux-firebase/blob/master/src/utils/storage.js#L78-L91). I'm glad you pointed this out because it seems that the promise is running on result of the meta set promise when it should go before it so that can be included in the meta (exactly as you are saying).

The change should be easy enough to make, but it would be nice to capture this functionality with a unit test as well while we are at it. Thanks for reporting.

PranavPanchal:
Hey, I am carrying over from where we last left off from this thread: #480 

I am using an async action. The following is the code:

`export const uploadToFirebase = file => async (
  dispatch,
  getState,
  { getFirebase, getFirestore }
) => {
  const fileName = cuid();
  const firebase = getFirebase();
  const firestore = getFirestore();

  const path = "files";
  const options = {
    name: fileName
  };

  try {
    // upload the file to firebase storage
    let uploadedFile = await firebase.uploadFile(path, file, null, options);

    console.log(uploadedFile.uploadTaskSnapshot.ref.getDownloadURL());

    // get url of the file
    let downloadURL = await uploadedFile.uploadTaskSnapshot.downloadURL;

    // add the new files to files collection
    await firestore.add(
      {
        collection: "files"
      },
      {
        name: fileName,
        url: ""
      }
    );
  } catch (error) {
    console.log(error);
  }
};`

The screenshot that I posted in the previous comment was from console logging the file that was uploaded to firebase storage.

`    console.log(uploadedFile.uploadTaskSnapshot.ref.getDownloadURL());`

[...]

Issue #480, opened 2018-05-30
bug(FileMeta): uploadFile errors out trying to access 0th item of undefined (prevents reading/writing file meta info/file URL)


nasimoyz:
**What is the current behavior?**

When attempting to upload an image using `uploadFile` or `uploadFiles` the following error occurs:
![image](https://user-images.githubusercontent.com/1500565/40692681-29f6b51a-6381-11e8-922e-0a40cfbf9f5b.png)

The file uploads, but fails to write metadata or subsequently update redux state: `firebase.data[filesPath]` is null

Error occurs in `utils/storage.js` function `writeMetadataToDb` when trying to access 0th item of downloadURLs, where downloadURLs is undefined (property does not exist on `uploadTaskSnapshot`)

<img width="708" alt="screen shot 2018-05-29 at 8 36 44 pm" src="https://user-images.githubusercontent.com/1500565/40692527-5bff898e-6380-11e8-89f0-dc941edb1a53.png">

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via [codesandbox](https://codesandbox.io/) or similar.**

```
...
this.props.firebase.uploadFile(filesPath, file, filesPath);
...
compose(
  firebaseConnect([ filesPath ]),
)(FormComponent);
```

**Which versions of dependencies, and which browser and OS are affected by this issue? Did this work in previous versions or setups?**

```
"firebase": "^5.0.4",
"react-redux-firebase": "^2.1.4",
```

I would downgrade firebase, but if conflicts with the firestore requirements.

prescottprue:
@nasimoyz Thanks for reporting. For now, it should be easy to just not include it if doesn't exist. It looks like to keep the results of the method the same, we may have to add an internal [`getDownloadURL`](https://firebase.google.com/docs/reference/js/firebase.storage.Reference#getDownloadURL) call.

prescottprue:
As mentioned by @harvitronix in #483, a current work around for this is:

```js
this.props.firebase.uploadFile(FILES_PATH, file)
  .then(ret => ret.uploadTaskSnapshot.ref.getDownloadURL())
  .then(url => console.log(url));
```

To keep the resulting metadata the same, I am going to change the code within [the v2.1.5 PR](https://github.com/prescottprue/react-redux-firebase/pull/481) to call the `getDownloadURL` method if the download URL is not already on the response (i.e. when using Firebase JS SDK that is v5).

[...]

PR #533, opened 2018-09-07
#487 fix(storage): fix downloadURL error in file upload metadata

jpierce42:
### Description
Fixes errors I ran into when trying to save file metadata to Firestore on upload:

- Async-related error: "ReferenceError: regeneratorRuntime is not defined" is fixed by installing babel/polyfill:  [https://github.com/babel/babel/issues/5085](https://github.com/babel/babel/issues/5085)

- Syntax for returning `downloadURL` from the fileMetadataFactory has changed; updated in docs.

Also including a small fix to the 'Getting Started' docs, which I encountered when setting up a sample project to test.

### Check List

- [x] All tests passing

(I did not add any tests, if needed please let me know.)

### Relevant Issues
#480 
#487 
#488

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "redux-react-firebase"
   ],
   "dependencies": {
+    "@babel/polyfill": "^7.0.0",
     "hoist-non-react-statics": "^2.3.1",
     "lodash": "^4.17.10",
     "prop-types": "^15.6.1"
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     "redux-react-firebase"
   ],
   "dependencies": {
+    "@babel/polyfill": "^7.0.0",
```
prescottprue: Maybe that this addition was needed due to node version or `babel-preset-env` settings needing to be changed since it hasn't been needed until now? Not sure.
Either way though, we can leave it and we can investigate more before it goes to `latest`.

PR #589, opened 2018-12-16
Fix context to Consumer instead of Provider

Rizzet:
### Description
Wrongly used ReactReduxFirebaseContext.Provider instead of ReactReduxFirebaseContext.Consumer generates an error from React when firebaseConnect() is used.
While firestoreConnect() is using ReactReduxFirebaseContext.Consumer this is probably only a typo.

### Check List

- [X] All tests passing


prescottprue:
That must have been changed during a cleanup. Thanks for reporting.

prescottprue:
Released in [v3.0.0-alpha.2](https://github.com/prescottprue/react-redux-firebase/releases/tag/v3.0.0-alpha.2).

PR #613, opened 2019-01-10
Fix issue with reserved prop names inside firebaseConnect

theashguy:
### Description

Prior to this commit firebaseConnect used the prop 'firebase' to receive the firebase app reference. This meant that if another prop called firebase (for instance, from the default reducer configuration) was supplied to a firebaseConnect wrapped component it would clobber the variable. This would result in either unexpected data in the store, or alternatively the inability for developers to nest firebaseConnect wrapped components.

An example case by @iamthefox has been uploaded here:
https://github.com/iamthefox/react-redux-firebase/blob/next-nested-issues/examples/complete/nested-issue/src/Home.js#L30

This commit renames the internal reference, and also checks to see if reserved words are passed through and throws an error with a useful message that can help developers debug if they accidentally provide one of these reserved props.

Note: This same issue *could* exist for firestore as well, however I don't have any examples to test with.

### Check List

- [ x ] All tests passing
- [ x ] Docs updated with any changes or examples if applicable
- [ x ] Added tests to ensure new feature(s) work properly

### Relevant Issues
Raised in gitter, example provided in description.


theashguy:
Fixed @prescottprue, have a look and let me know if you want anything changed and / or a different approach entirely. Tried to make it resilient to this class of errors in the future. 

Unfortunately we can't test for what the developer is going to pass into a connected component in build (in fact in some cases the developer might not even know off the top of their head), but we can prevent clashes during execution and raise a helpful error, plus make the internals obscure enough that they wont get hit.

src/firebaseConnect.js
```diff
@@ -89,23 +97,40 @@ export const createFirebaseConnect = (storeKey = 'store') => (
...
+    // this issue.
+    const clashes = Object.keys(props).filter(
+      k => !!RESERVED_PROPS.find(r => r === k)
+    )
+
+    if (clashes) {
+      throw new Error(
+        `Supplied prop/s "${clashes.join(
+          '", "'
+        )}" are reserved for internal firebaseConnect() useage.`
```
prescottprue: Great error message, but it should be "usage" right?

prescottprue:
Really awesome to see the fix and such a comprehensive PR/explanation around it, thanks! I'll merge after testing it a bit

src/firebaseConnect.js
theashguy: Haha, correct. I will fix.

src/firebaseConnect.js
```diff
@@ -89,23 +97,40 @@ export const createFirebaseConnect = (storeKey = 'store') => (
...
-  )
+  const FirebaseConnect = props => {
+    // Check that reserved props are not supplied to a FirebaseConnected
+    // component and if they are, throw an error so the developer can rectify
+    // this issue.
+    const clashes = Object.keys(props).filter(
+      k => !!RESERVED_PROPS.find(r => r === k)
+    )
+
+    if (clashes) {
```
iamthefox: should be if (clashes.length), !![] will resolve as true

src/createFirebaseInstance.js
```diff
@@ -22,6 +22,8 @@ export default function createFirebaseInstance(firebase, configs, dispatch) {
     firebase.database.enableLogging(configs.enableLogging)
   }
 
+  console.log('Creating firebase instance', firebase)
```
iamthefox: @theashguy this is causing CI to fail due to eslint rules. 

If you need this log here might be worth disabling eslint check for this line otherwise you can get rid of it

src/firebaseConnect.js
theashguy: Correct, I have fixed this already.

prescottprue:
This is causing some issues with the usage of the `firebase` prop downstream. I do like the idea of handling the case of this prop being passed and throwing, but I think renaming or disabling the `firebase` prop will have to be an option, at least for now.

Going to keep in the throwing, but add back the `firebase` prop.

PR #625, opened 2019-02-05
Add types to reducers and getFirebase, fix ReactReduxFirebaseProviderProps

idanlo:
### Description
I added return types for the reducers (firebase and firestore), these are not yet finished, some of properties are still `any`, these come from my experience working with `react-redux-firebase` and `redux-firestore` so it may be incorrect and I'm happy to be corrected.

There was an error with `ReactReduxFirebaseProviderProps` which was generic but wasn't given any generic type so I gave it `any` _for now_.

In my app I noticed that the `getFirebase` function returns the type `typeof firebase` which is the entire namespace, I'm not sure about this so fix me if I'm wrong.


rscotten:
@idanlo can you please add in the work that I did on this?

https://github.com/rscotten/react-redux-firebase/blob/v3.0.0-alpha.6-upgrade-ts-definitions/index.d.ts

idanlo:
> @idanlo can you please add in the work that I did on this?
> 
> https://github.com/rscotten/react-redux-firebase/blob/v3.0.0-alpha.6-upgrade-ts-definitions/index.d.ts

@rscotten Working on it. are both `firebase` & `@firebase/*` dependencies?

idanlo:
@rscotten I added some types. I don't understand some of your types because you aren't using them anywhere. some of the things you typed can be imported directly from firebase, for example `ReduxFirestoreApi` with all the `get, set, add` methods can be imported from firebase, they already have types for that. Also you aren't using those types in the file, for example setting the return type for `withFirebase`

idanlo:
regarding https://github.com/prescottprue/react-redux-firebase/pull/625/commits/20c6896e0e3eb2e5a7a7677c8b594620b02f6cbd 

I am not sure about the return types for `withFirebase` and `withFirestore` because the `WithFirebaseProps` and `WithFirestoreProps` are not the only props that are passed to the returned component, there are also the props for the component that the user wrote.

prescottprue:
Hey, whoops, didn't catch this one initially and merged the other typings PR. Let me look at this in a little bit. Thanks for all of your work @idanlo !

rscotten:
Hey, @idanlo 

@prescottprue just merged in type changes into the next branch.

These types look like they were taken from an old version of mine.  But it has a lot of stuff I missed.  Should I merge them in and create a new PR, or does @idanlo want to do it?

idanlo:
You can do it, @rscotten

rscotten:
@idanlo  Great.  I'll do it now.  Please close this PR.

PR #664, opened 2019-03-31
changed attachAuthIsReady default to false

djpeach:
it is not true by default

### Description

just a small docs thing, but it was annoying when what I was doing did not match what the docs indicated. No use leaving it wrong.

### Check List

- [x ] All tests passing
- [x ] Docs updated with any changes or examples if applicable
- [x ] Added tests to ensure new feature(s) work properly


prescottprue:
Great catch, thanks for making the PR!

Issue #667, opened 2019-04-04
bug(types): wrong types for firebaseConnect


lazarljubenovic:
**Do you want to request a *feature* or report a *bug*?**  

Bug.

(If this is a *usage question*, please **do not post it here**—post it on [gitter](https://gitter.im/redux-firebase/Lobby). If this is not a “feature” or a “bug”, or the phrase “How do I...?” applies, then it's probably a usage question.)

**What is the current behavior?**

Writing

```ts
const firebaseConnected = firebaseConnect(() => [
  { path: 'gameScores' },
])
```

causes

```text
TypeScript error: Argument of type '() => { path: string; }[]' is not assignable to parameter of type 'string[] | mapper<{}, string[]> | undefined'.
  Type '() => { path: string; }[]' is not assignable to type 'mapper<{}, string[]>'.
    Type '{ path: string; }[]' is not assignable to type 'string[]'.
      Type '{ path: string; }' is not assignable to type 'string'.  TS2345
```

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via [codesandbox](https://codesandbox.io/) or similar.**

I think it's trivial, but I can make a Blitz if needed.

**What is the expected behavior?**

I should be able to use the extended (non-string) notation when connecting.

**Which versions of dependencies, and which browser and OS are affected by this issue? Did this work in previous versions or setups?**

Tested with a local build of the current tip of v3 branch, as well as the `@next` package from npm; fails on both with the same error. 

Gonna try fixing it now, but in case I fail, hopefully someone else can chime in. :sparkles:

prescottprue:
@lazarljubenovic I have noticed that the types on the `next` branch aren't perfect as well. I have started updating some of them, so I will check into this one. @rscotten has also been a huge help in updating them, so one of us should be able to get to it soon.

lazarljubenovic:
@prescottprue Thanks for the update! I've change the .d.ts locally like this:

```ts
export interface FirebaseConnectQueryObject {
  path: string
  type?: 'value' | 'once' | 'child_added' | 'child_removed' | 'child_changed' | 'child_moved'
  queryParams?: string[]
}

export type FirebaseConnectQuery = (FirebaseConnectQueryObject | string)[]

/**
 * React HOC that attaches/detaches Firebase Real Time Database listeners on mount/unmount
 */
export function firebaseConnect<ProfileType, TInner = {}>(
[...]

PR #668, opened 2019-04-04
fix(types): arguments for "firebaseConnect"

lazarljubenovic:
### Description

Previous, only strings were allows in `firebaseConnect` based on types.

```ts
const firebaseConnected = firebaseConnect(() => [
  'gameScores',
])
```

This PR enables using the alternative as well.

```ts
const firebaseConnected = firebaseConnect(() => [
  { path: 'gameScores' },
])
```

### Check List

- [x] All tests passing

We should probably eventually force the TS example to contain lots of different things for the sake of demonstration _and_ testing types, but seems like it's still in a very basic state anyway, so I don't think that adding these now should be a part of this PR; it will just postpone everything too long since looks like a lot of things are missing in the example anyway.

### Relevant Issues

* #667

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -448,11 +448,19 @@ export interface WithFirebaseProps<ProfileType> {
     }
 }
 
+export interface FirebaseConnectQueryObject {
+  path: string
+  type?: 'value' | 'once' | 'child_added' | 'child_removed' | 'child_changed' | 'child_moved'
+  queryParams?: string[]
+}
+
+export type FirebaseConnectQuery = (FirebaseConnectQueryObject | string)[]
+
 /**
  * React HOC that attaches/detaches Firebase Real Time Database listeners on mount/unmount
  */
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -448,11 +448,19 @@ export interface WithFirebaseProps<ProfileType> {
     }
 }
 
+export interface FirebaseConnectQueryObject {
```
prescottprue: Nice!

prescottprue:
I have been wanting to add the configs for the for a while, so thanks for making the PR! There are some [other updates I have on the `types-overhaul` branch](https://github.com/prescottprue/react-redux-firebase/tree/types-overhaul), so we can combine this fix with those.

lazarljubenovic:
Oh, didn't see that branch. I'll check it out tomorrow and rebase.

lazarljubenovic:
Gonna close this in favor of https://github.com/prescottprue/react-redux-firebase/pull/669; re-did work on the other branch.

PR #669, opened 2019-04-05
Types overhaul

lazarljubenovic:
### Description

See #668.

I added back `<ProfileType>` in a few places so the file would be without errors; we can clean this up if we want later.

### Check List

- [x] All tests passing

### Relevant Issues

* #667


prescottprue:
Great work - I was starting to change that since it didn't read correctly at all (wasn't sure why ProfileType would be passed to `withFirebase`)

PR #685, opened 2019-04-17
fix(examples): updated scripts in material example to match Readme

Shalinit3:
### Description
Will change docs sections for the build script.
Will add 'start:build' script that builds the application and then serves at localhost:3000 using firebase serve.

Readme of examples/material: https://github.com/prescottprue/react-redux-firebase/tree/master/examples/complete/material#getting-started

### Check List

- [x] All tests passing

### DIFF
--- a/examples/complete/material/.gitignore
+++ b/examples/complete/material/.gitignore
@@ -8,6 +8,6 @@
 **/node_modules
 
 # React App
-**/dist
+**/build
 src/config.js
 .env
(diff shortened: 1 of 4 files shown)


prescottprue:
Instead of moving the folder I think it may be better to just adjust the docs to match since that is how it is [within generator-react-firebase](https://github.com/prescottprue/generator-react-firebase). To fix it working with the emulator, we should just be able to change this line: https://github.com/prescottprue/react-redux-firebase/blob/master/examples/complete/material/firebase.json#L9 - I believe this was around from an old run of generator-react-firebase.

Thanks for reporting!

Shalinit3:
@prescottprue done the changes!

PR #695, opened 2019-04-30
fix(tests): replace instances of "to.equal" with "to.be" in tests - @Shalinit3

Shalinit3:
Will replace to.equal(null) with to.be.null, to.equal(undefined) with to.be.undefined and to.equal(true) with to.be.true in test cases.

### Check List

- [x] All tests passing


prescottprue:
Great catch! Thanks for the PR 😄

I have used those on some other more recently written tests, especially for other tools, but never went back to update these ones.

PR #696, opened 2019-04-30
fix(core): add lint settings for react version to prevent warning while running lint - @Shalinit3

Shalinit3:
Will remove the warning message(Warning: React version not specified in eslint-plugin-react settings. See https://github.com/yannickcr/eslint-plugin-react#configuration.) while running lint.

### Check List

- [x] All tests passing

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,12 @@ parser: babel-eslint
 extends: [standard, standard-react, prettier, prettier/react]
 plugins: [babel, react, prettier]
 
+settings: {
+  react: {
+    version: 'detect'
+  }
+}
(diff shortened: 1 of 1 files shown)


prescottprue:
Great work, thanks for the fix!

Not sure if I mentioned before, but the material example generated using [generator-react-firebase](https://github.com/prescottprue/generator-react-firebase). When using that to start a project, it uses [the JS formatted `.eslint.js`](https://github.com/prescottprue/generator-react-firebase/blob/master/examples/react-firebase-redux/.eslintrc.js), so maybe in the future we can do that here too.

PR #701, opened 2019-05-17
fix(profile): pass updateProfile options to action

cruzdanilo:
### Description
fix options not being passed to updateProfile action


prescottprue:
Going to make a separate commit to avoid conflicts with master as I  put this into an upcoming `latest` release. Will also get that out to next sometime soon.

prescottprue:
Included in [v3.0.0-alpha.13 release](https://github.com/prescottprue/react-redux-firebase/releases/tag/v3.0.0-alpha.13). The change is also planned to go out into latest with v2.3.0. Thanks for the PR!

PR #725, opened 2019-07-08
fix(providers): prevent unwanted re-initialization when parent component updates - @illuminist

illuminist:
### Description
I wrapped an extended firebase instance inside react useMemo hook in order to prevent `extendedFirebaseInstance.initializeAuth()` from being called for second time when parent component updated. And it should has improved performance since it wouldn't call `createFirebaseInstance` again to create a new and change firebase instance in the `ReactReduxFirebaseContext` which would cause  update in every component that depends on firebase instance from said context.

### Check List

- [x] All tests passing

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "prop-types": "^15.6.2"
   },
   "peerDependencies": {
-    "react": "^16.4.0"
+    "react": "^16.8.0"
   },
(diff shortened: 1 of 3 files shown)


prescottprue:
Great work! 🎆 Really excited to see the hooks continuing to move forward

PR #730, opened 2019-07-18
Fix `useFirestoreConnect` type

illuminist:
### Description
`useFirestoreConnect` type was set to accept array of query but it should accept only single query per hook.

- [x] All tests passing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -668,9 +668,9 @@ export function useFirestore(): ExtendedFirestoreInstance
  */
 export function useFirestoreConnect<TInner>(
   connect?:
-    | mapper<TInner, (string | ReduxFirestoreQuerySetting)[]>
-    | ReduxFirestoreQuerySetting[]
-    | string[]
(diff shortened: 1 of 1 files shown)


prescottprue:
Remind me again, Wwat was the reasoning for just the one query and not sticking to the original API? Was it for memoization?

Thanks for the PR

illuminist:
Mainly because I wanted to take adventage `useEffect`. Because `useEffect` will automatically set and unset a listener if it's dependencies has changed, in this case, the query.

In the original API, it uses `componentDidUpdate` lifecycle to compare between an old and a new query. But it's unable to compare the old query using `useEffect` because it cannot access the old query directly without some workaround which I didn't know at the time. Therefore I made to unset a listener every time the query has changed. This comes to the problem with multi query connector where one query might be updated but will cause every listener in a connector to be reset, causing more unnecessary read to firestore.

I just found a new way to compare old and new state using hook, which might work same way as the old implementation. I will try that first.

prescottprue:
Nice!! Thanks for posting the details of your findings

PR #781, opened 2019-10-21
fix(types): fix firestore.delete return type - @rscotten

rscotten:
### Description

Type fix that changes the return type in line 482 in `index.d.ts` from 

```typescript
delete: (docPath: string | ReduxFirestoreQuerySetting) => void
```

to
```typescript
delete: (docPath: string | ReduxFirestoreQuerySetting) => Promise<void>
```

### Check List

- [x] All tests passing


rscotten:
@prescottprue I branched off master instead of next so I could use the latest 3.0.2 release.  Looks like next is still on 3.0.0.

prescottprue:
Thanks for the PR! The v3.0.0 version is out in `latest`, so I am going to switch this to be against the next upcoming patch instead of `next`

Issue #788, opened 2019-11-04
feat(example): Examples of testing components using RRF Hooks? (Jest/React-Testing-Library etc.)


dsgriffin:
Hi all, wondering if there are any examples of how to test components/providers that use React Redux Firebase hooks? Specifically using Jest/React Testing Library but other libraries are welcome. Thanks!

harish-aka-shivi:
Looking into it. 🙂

prescottprue:
Thanks @harish-aka-shivi!

PR #824, opened 2020-01-05
test(example): add test cases to simple firestore example

harish-aka-shivi:
This commit adds test cases to firestore example to let users know how to test the components which are using RRF hooks. 

Resolves: #788 

### Check List

- [x] All tests passing

### DIFF
--- a/examples/complete/firestore/src/Todos.test.js
+++ b/examples/complete/firestore/src/Todos.test.js
@@ -23,7 +23,7 @@ describe('Should test the Todos Component', () => {
   })
 
   it('show empty text when the todo list is empty', () => {
-    let todos = []
+    const todos = []
     useSelector.mockReturnValue(todos)
 
     act(() => {
@@ -35,7 +35,7 @@ describe('Should test the Todos Component', () => {
   })
 
   it('should have one TodoItem in the list', () => {
-    let todos = [
+    const todos = [
       {
         id: 'todoId',
         name: 'Sample todo',
--- a/examples/complete/firestore/src/Todos.test.js
+++ b/examples/complete/firestore/src/Todos.test.js
@@ -6,9 +6,8 @@ import Todos from './Todos'
 
 jest.mock('react-redux')
 
-describe('Should test the sample component', () => {
+describe('Should test the Todos Component', () => {
   let container = null
-  const expectedTodos = []
 
   beforeEach(() => {
     // setup a DOM element as a render target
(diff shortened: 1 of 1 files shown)


harish-aka-shivi:
Please review the code and check if the current test case is fine. 
Also, should I add the test cases for all the components using the RRF hooks in this example?

examples/complete/firestore/src/Todos.test.js
```diff
@@ -0,0 +1,50 @@
...
+
+  afterEach(() => {
+    // cleanup on exiting
+    unmountComponentAtNode(container)
+    container.remove()
+    container = null
+  })
+
+  it('show empty text when there is no text', () => {
+    useSelector.mockReturnValue(expectedTodos)
```
prescottprue: Can we make `expectedTodos` within the scope of the test instead of global to all tests?

examples/complete/firestore/src/Todos.test.js
```diff
@@ -0,0 +1,50 @@
...
+    expect(container.textContent).toBe('Todo list is empty')
+  })
+
+  it('should have on TodoItem when there is one to do', () => {
+    let moreTodo = {
+      id: 'todoId',
+      name: 'Sample todo',
+      text: 'This is for testing todos'
+    }
+    expectedTodos.push(moreTodo)
```
prescottprue: Same here, if we make a new array within the scope of this test, it can contain the new todo and won't share variables/scope with other tests

examples/complete/firestore/src/Todos.test.js
```diff
@@ -0,0 +1,50 @@
...
+    useSelector.mockReturnValue(expectedTodos)
+    act(() => {
+      render(<Todos />, container)
+    })
+    const listItems = container.querySelectorAll('li')
+        // this is not a real id, so only text which can be visible is Delete
```
prescottprue: Not sure I understand this comment - what ID is it referring to? And wouldn't `Delete` be within the button in that list item?

harish-aka-shivi: `TodoItem` in this example is using the `id` from the props to fetch the Todo from the firebase server. For the mock purpose, we have given a fake Id, which will not return any todo as it does not exist, only `delete` text will be shown. But we can imply that `TodoItem` is being rendered.

examples/complete/firestore/src/Todos.test.js
harish-aka-shivi: Sure, I will change it.

examples/complete/firestore/src/Todos.test.js
harish-aka-shivi: I will change it.

examples/complete/firestore/src/Todos.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+  afterEach(() => {
+    // cleanup on exiting
+    unmountComponentAtNode(container)
+    container.remove()
+    container = null
+  })
+
+  it('show empty text when the todo list is empty', () => {
+    let todos = []
```
prescottprue: This can be `const`

examples/complete/firestore/src/Todos.test.js
```diff
@@ -0,0 +1,56 @@
...
+    act(() => {
+      render(<Todos />, container)
+    })
+
+    container.querySelectorAll('li')
+    expect(container.textContent).toBe('Todo list is empty')
+  })
+
+  it('should have one TodoItem in the list', () => {
+    let todos = [
```
prescottprue: This can be const

examples/complete/firestore/src/Todos.test.js
prescottprue: Gotcha - thanks for the explanation

prescottprue:
If we just change the `let` instances to `const` then this should be good to ship

examples/complete/firestore/src/Todos.test.js
harish-aka-shivi: done

Issue #670, opened 2019-04-05
Providing type for the schema


lazarljubenovic:
**Do you want to request a *feature* or report a *bug*?**  Feature

(If this is a *usage question*, please **do not post it here**—post it on [gitter](https://gitter.im/redux-firebase/Lobby). If this is not a “feature” or a “bug”, or the phrase “How do I...?” applies, then it's probably a usage question.)

**What is the current behavior?**

No type info for the schema of used database.

**What is the expected behavior?**

Moar typez.

# Basic idea

The library provides an empty interface:

```ts
export interface FirebaseRtdbSchema { }
```

Developer has his own types for the schema. Not related to the lib in any way, he can do whatever he wants.

<details>

```ts
// some helpers
type RtdbUid<T> = string & { __UID_FOR__: T }
type FirebaseRecord<T> = Record<RtdbUid<T>, T>
type FirebaseKvPair<T> = { key: RtdbUid<T>, value: T }
type FirebaseKvPairs<T> = Array<FirebaseKvPair<T>>

interface RtdbUser {
  avatarUrl: string
  displayName: string
  email: string
}

interface RtdbGameScore {
  score: number
  size: number
  speed: number
  user: RtdbUid<RtdbUser>
  userDisplayName: string
}

interface RtdbSchema {
  users: FirebaseRecord<RtdbUser>
  gameScores: FirebaseRecord<RtdbGameScore>
}
```

</details>

Developer creates a `.d.ts` file where he extends the existing typings by merging in his own schema:

```ts
// make react-redux-firebase aware of the schema, globally
import 'react-redux-firebase'
declare module 'react-redux-firebase' {
  export interface FirebaseRtdbSchema extends RtdbSchema { }
}
```

Various methods and properties in the library can now use `FirebaseRtdbSchema` to bring more type safety by having a generic `fnName<T extends Record<string, any> = FirebaseRtdbSchema>`. If the developer doesn't provide anything, the globally declared schema is used; overriding mechanism allows for having multiple databases and such for advanced use cases.

Leveraging this means that the developer gets things like the following autocompletion.

![image](https://user-images.githubusercontent.com/7661457/55623901-948c8180-57a4-11e9-9967-e6eaf56dfa1f.png)

# Nested keys

When the lib requires specifying a path such as `foo`, `foo.bar` or `foo/bar`, we can provide a simple `.join('/')` function which has some magic types:

```ts
function chainOfKeys <
  T extends Record<string, any> = Schema,
[...]

PR #826, opened 2020-01-07
feat(types): add support for typed db schema

vicrac:
### Description
For now, loose typings for state make it unsafe to use with typescript. I've added additional type parameter to state, defining `Schema` for database. Consider:
```ts
interface Todo {
  description: string;
}

// create schema for the DB
interface DBSchema {
  todos: Todo;
}

const rootReducer = combineReducers({
  firebase: makeFirebaseReducer<DBSchema>()  // new API for passing type parameter
})

type AppState = ReturnType<typeof rootReducer>;
``` 
Now if we create a selector...
```ts
export const firebaseSelector = (state: AppState) => state.firebase;

export const allItemsSelector = createSelector(
  firebaseSelector,
  firebase => firebase.data.todos
);
```
we are getting autocompletion!:
image

... and proper inner type!
image

Of course, I am open to suggestions, but I literally made the smallest possible change that allowed me to get these features (and not to crash types).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1132,7 +1132,9 @@ export namespace FirebaseReducer {
     profile: Profile<ProfileType>
     authError: any
     data: { [T in keyof Schema]: Record<string, Schema[T]> }
-    ordered: Ordered<any>
+    ordered: {
+      [T in keyof Schema]: Array<{ key: string; value: Schema[T] }>
+    }
     errors: any[]
     isInitializing: boolean
     listeners: Listeners
--- a/index.d.ts
+++ b/index.d.ts
@@ -805,10 +805,15 @@ export function firebaseConnect<ProfileType, TInner = {}>(
  * @param action.data - Data associated with action
  * @see https://react-redux-firebase.com/docs/api/reducer.html
  */
-export function firebaseReducer<UserType>(
-  state: any,
-  action: any
-): FirebaseReducer.Reducer<UserType>
+export function firebaseReducer<
+  Schema extends Record<string, Record<string | number, string | number>>,
+  UserType
+>(state: any, action: any): FirebaseReducer.Reducer<Schema, UserType>
+
+export function makeFirebaseReducer<
+  Schema extends Record<string, Record<string | number, string | number>>,
+  UserType = {}
+>(): (state: any, action: any) => FirebaseReducer.Reducer<Schema, UserType>
 
 /**
  * React HOC that attaches/detaches Cloud Firestore listeners on mount/unmount
@@ -1119,11 +1124,14 @@ export interface Data<T extends FirestoreTypes.DocumentData> {
 }
 
 export namespace FirebaseReducer {
-  export interface Reducer<ProfileType = {}> {
+  export interface Reducer<
+    Schema extends Record<string, Record<string | number, string | number>>,
+    ProfileType = {}
+  > {
     auth: AuthState
     profile: Profile<ProfileType>
     authError: any
-    data: Data<any | Dictionary<any>>
+    data: { [T in keyof Schema]: Record<string, Schema[T]> }
     ordered: Ordered<any>
     errors: any[]
     isInitializing: boolean


prescottprue:
Amazing work! Thanks for the PR! I think this is part of what what was being described in #670 

Since this doesn't seem to be breaking in anyway we should be able to get this out in a new minor version

vicrac:
That's great :smile: I gave it a bit thought and here are things that we should consider (**but note it's not necesarrily to be implemented now. I think these types should be rather released as soon as possible, feedback should be gathered and then they should be gradually improved**):
1. Doing the same kind of types for `firestore`
2. I realized currently my solution works only for not-nested objects, due to `Record<string | number, string | number>` given as type. There should be developed a generic type for any kind of data that can be stored in database (it will probably be a recursive type, sth like `type DBObject = Record<string | number, string | number | DBObject>`) if there's no such yet.
3. What if schema for DB contains not only collections, but also some "unique" object (e.g. for storing denormalized counters)? I mean
```ts
{
  todos: { ... },
  counters: {
    todos: 15
  }
}
```
it should be reflected in schema so it doesn't produce a `Record` from it. I thought of sth like:
```ts
import { Collection } from 'react-redux-firebase';

interface Counters {
  todos: number;
}

interface Schema {
  todos: Collection<Todo>;
  counters: Counters;
}
```

and then branching on it using most probably conditional types.
4. I've also seen people doing some nesting for joins' purposes, e.g
```ts
{
  users: { 
    1: { name: "John Doe" }
  },
  todos: {
    1: {  // user's 1 todos
      2: {
        description: "Some todo"
      }
    }
  }
}
```
(then retrieving todos for user 1 is simpler). How should it be handled? How is it currently handled by implementation itself?

vicrac:
Guys, any conclusions on above? I'd love to have this feature in my project :)
@prescottprue

prescottprue:
Love the idea you are mentioning above, but agree it can come at a later time.

Going to point this to v3.1.0 and get it released - I'm thinking we will also want to add something about this to the documentation for other folks to get the same benefits. Is the `makeFirebaseReducer` function necessary or have you found a way for the type to be passed with the default reducer import and assignment?

vicrac:
@prescottprue Great. Unfortunately, `makeFirebaseReducer` is necessary to pass type parameter - Typescript doesn't have polymorphic literals, so you can't do
```ts
combineReducers({
  firebase: firebaseReducer<DBSchema>  // <-- won't compile
})
```

Also, I don't really have free time to polish it nor write proper documentation, but in case of any questions regarding Typescript/anything else I'd be happy to help :+1:

rscotten:
I love the idea and intention here, but this PR is incomplete, and its release a little premature because it broke my typed code.

1. It only defines a Schema for Firebase Realtime DB, not Firestore.
2. It really needs some documentation on how to implement.  Like, how do I type this?

```typescript
import { combineReducers } from 'redux';
import { firestoreReducer } from 'redux-firestore';
import { firebaseReducer } from 'react-redux-firebase';

interface User { };

interface Schema { };

interface RootState {
  // previously
  firebase: FirebaseReducer.Reducer<User>;
  // change to
  firebase: FirebaseReducer.Reducer<Schema, User>;
  firestore: FirestoreReducer.Reducer;
}

combineReducers<RootState>({
    firebase: firebaseReducer,
    firestore: firestoreReducer,
  });
```

I admit figuring out this change is not too difficult for a Typescript developer to figure out, but documentation with this release would have saved me some debugging time.

rscotten:
@prescottprue @vicrac 
I love the idea and intention here, but this PR is incomplete, and its release a little premature because it broke my typed code.

1. It only defines a Schema for Firebase Realtime DB, not Firestore.
2. It really needs some documentation on how to implement.  Like, how do I type this?

```typescript
import { combineReducers } from 'redux';
import { firestoreReducer } from 'redux-firestore';
import { firebaseReducer } from 'react-redux-firebase';

interface User { };

interface Schema { };

interface RootState {
  // previously
  firebase: FirebaseReducer.Reducer<User>;
  // change to
  firebase: FirebaseReducer.Reducer<Schema, User>;
  firestore: FirestoreReducer.Reducer;
}

combineReducers<RootState>({
    firebase: firebaseReducer,
    firestore: firestoreReducer,
  });
```

I admit figuring out this change is not too difficult for a Typescript developer to figure out, but documentation with this release would have saved me some debugging time.

rscotten:
Also, `Schema` should have been made the second generic parameter to `FirebaseReducer.Reducer` not replacing `UserType` as the first.  Thi is going to cause everyone's code to break.

prescottprue:
@rscotten Thanks for pointing out - I'm open to a PR if you get a chance

w0nche0l:
Yeah this was definitely an annoying breaking change to figure out, please put it in the patch notes or something?

prescottprue:
@w0nche0l The support for schema was mentioned in the release notes, but the break of types was not intended as part of the change. Instead of noting the workaround I am making the update mentioned by @rscotten (i.e. having Schema be the second generic type). Also, I have been working on updating the typescript example and adding a page to the recipes section of the docs explaining how to pass a schema type when using Typescript.

As for the schema handling for Firestore, I'll have to get to that when I can unless someone else wants to make a PR

index.d.ts
```diff
@@ -1119,12 +1124,17 @@ export interface Data<T extends FirestoreTypes.DocumentData> {
 }
 
 export namespace FirebaseReducer {
-  export interface Reducer<ProfileType = {}> {
```
macinjoke: This is breaking change. The parameter order was changed.

prescottprue: Good call out. Yeah, this was in an attempt to have the feature work as originally intended

Issue #838, opened 2020-01-19
Logout causes rejection error


jhefreyzz:
**Do you want to request a *feature* or report a *bug*?**   bug

(If this is a *usage question*, please **do not post it here**—post it on [gitter](https://gitter.im/redux-firebase/Lobby). If this is not a “feature” or a “bug”, or the phrase “How do I...?” applies, then it's probably a usage question.)

This issue relates to #529 which was closed.

**What is the current behavior?**
When a authenticated user logs out, it will also logout to same app open in other tab or windows. The behavior seems normal until a few seconds a rejection will be thrown.
`Unhandled Rejection (FirebaseError): Missing or insufficient permissions.`

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem via [codesandbox](https://codesandbox.io/) or similar.**
![issue](https://user-images.githubusercontent.com/20869829/72678083-857d8280-3add-11ea-8d2b-6449038f54e5.gif)

**What is the expected behavior?**
The app should not thrown this error (I'm not sure of it).

Security Rules:
```
rules_version = '2';
service cloud.firestore {
  match /databases/{database}/documents {
    match /{document=**} {
      allow read, write: if request.auth.uid != null;
    }
  }
}
```
Current config used:
```
const config = {
  userProfile: "users",
  useFirestoreForProfile: true, 
  enableLogging: false
};
```

**Which versions of dependencies, and which browser and OS are affected by this issue? Did this work in previous versions or setups?**
OS: Windows 10
Browser: Brave Browser Version 1.2.43
Dependencies: All packages are in the latest version

prescottprue:
Do you attach any listeners on the page which is loaded? Did it show what path the auth error is for? Are any errors written to redux state?

Seems like it may be the profile listener?

Can you provide this logic in a repo, codesandbox, or other similar tool so I can make sure I am fully replicating on my end?

prescottprue:
After searching through the issues it seems https://github.com/prescottprue/react-redux-firebase/issues/508 may be related or even the same issue

jhefreyzz:
Hi @prescottprue sorry take too long to reply.
I've created a repo https://github.com/jhefreyzz/react-redux-firebase-demo to replicate the issue mentioned above.

prescottprue:
Fix [released in v3.1.0](https://github.com/prescottprue/react-redux-firebase/releases/tag/v3.1.0) - thanks to @illuminist for the PR!

[...]

Issue #508, opened 2018-07-18
Error with profile listener: Missing or insufficient permissions. Error: Missing or insufficient permissions.


wallacepeng:
When i logout from firebase, i got below errors

Error with profile listener: Missing or insufficient permissions. Error: Missing or insufficient permissions.
    at new FirestoreError (index.cjs.js:346)
    at JsonProtoSerializer../node_modules/@firebase/firestore/dist/index.cjs.js.JsonProtoSerializer.fromRpcStatus (index.cjs.js:5258)
    at JsonProtoSerializer../node_modules/@firebase/firestore/dist/index.cjs.js.JsonProtoSerializer.fromWatchChange (index.cjs.js:5756)
    at PersistentListenStream../node_modules/@firebase/firestore/dist/index.cjs.js.PersistentListenStream.onMessage (index.cjs.js:13652)
    at index.cjs.js:13597
    at PersistentListenStream.<anonymous> (index.cjs.js:13573)
    at step (tslib.es6.js:91)
    at Object.next (tslib.es6.js:72)
    at tslib.es6.js:65
    at new Promise (<anonymous>)
    at Object.__awaiter (tslib.es6.js:61)
    at index.cjs.js:13569
    at index.cjs.js:13045

lhas2:
Hi @wallacepeng ! I'm receiving the same error. Did you find a solution for this?

wallacepeng:
@lhas2 This is caused by using user profile for login with redirection handling enabled. You can workaround it by setting UserProfile to null in the config. 
like 

const rrfConfig = {
  userProfile: null,
...
...
};

prescottprue:
Duplicate of #494. Error logs are just warnings (though they use console.error) and can be ignored as they don't impact functionality. Going to look into a fix this weekend.

wallacepeng:
thanks.

MoSwilam:
I am still getting the same exact error, and it can't be ignored as any reload prevents the app from reading information from firestore, any workaround yet ? thanks !

prescottprue:
@MoSwilam What version are you using? Have you tried the `logErrors: false` config option?

prescottprue:
@debaosuidecl `logErrors: false` should be passed as part of your config into the store enhancer ( `reactReduxFirebase`) when creating your store:

```js
const createStoreWithFirebase = compose(
  reactReduxFirebase(firebase, rrfConfig), // firebase instance as first argument
[...]

PR #840, opened 2020-01-22
fix(auth): remove stale profile listener on empty auth state change - @illuminist - #508, #838

illuminist:
### Description
If using `userProfile` and listening path is restricted with security rules, the profile listener will stay subscribed and cause permission error when user logged out. This initially can be prevented by using `firebase.logout` which has a call to unwatch user profile, however, there is some case that firebase will force user to logout, for example, logging out via other tab. This can't be prevented by using `firebase.logout` because `auth.onAuthStateChanged` is being trigger by firebase itself.

I have tested with pure firebase sdk, and found out that unwatching user profile inside of `auth.onAuthStateChanged` can fixed the issue from stale listener.

### Check List

- [x] All tests passing

### Relevant Issues
- https://github.com/prescottprue/react-redux-firebase/issues/838
- https://github.com/prescottprue/react-redux-firebase/issues/508

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -387,6 +387,8 @@ const handleAuthStateChange = (dispatch, firebase, authData) => {
       type: actionTypes.AUTH_EMPTY_CHANGE,
       preserve: config.preserveOnEmptyAuthChange
     })
+
+    unWatchUserProfile(firebase)
   } else {
     firebase._.authUid = authData.uid // eslint-disable-line no-param-reassign
 


prescottprue:
Amazing work! 👏 

Been wondering what might be causing those issues, and your explanation makes total sense! Thanks for the PR

illuminist:
There's already a few people having same issue as I gave them some advice in gitter, so I think this should be an improvement and should not break something as in any case I can think of. In more plus side, this will make user to be able to call either `firebase.logout()` and `firebase.auth().signOut()` without breaking something.

Tellefsonjr:
I couldn't find any threads for my particular issue, but this thread kept coming up - my issue was not with the profile listener, but with firestoreConnect() not clearing out the listeners after firebase.logout() method.

Changed from:
` export default compose(
    connect(mapStateToProps),
    firestoreConnect((state) => {
        ['collection']
    }),
  )(MyComponent)`

Changed to:
` export default compose(
    connect(mapStateToProps),
    firestoreConnect((state) => {
        return state.auth.isLoaded && state.auth.uid ? ['collection'] : []
    }),
  )(MyComponent)`

Not sure if this is intended, but really wished the firestoreConnect docs listed how to handle the listeners after logout :(

Issue #796, opened 2019-12-05
Warn if trying to use `firebase.initializeAuth` without importing 'firebase/auth'


illuminist:
**Do you want to request a *feature* or report a *bug*?**  

Feature

**What is the current behavior?**

If I forget to import 'firebase/auth' and call `firebase.initializeAuth()` or use ReactReduxFirebaseProvider with `initializeAuth` prop, the thunk action will exit without explanation which make it hard to track down what is missing.

https://github.com/prescottprue/react-redux-firebase/blob/master/src/actions/auth.js#L455-L458 

**What is the expected behavior?**

It should log warning in console, or dispatch auth error action, or some way to notify a developer that they forget to import 'firebase/auth'

**Which versions of dependencies, and which browser and OS are affected by this issue? Did this work in previous versions or setups?**

react-redux-firebase@3.0.5

markoelez:
@illuminist  @prescottprue  Could I try tackling this issue? I am looking to get some practice :)

PR #841, opened 2020-01-22
Throw error for auth init if firebase/auth isn't imported

illuminist:
### Description
https://github.com/prescottprue/react-redux-firebase/issues/796

I make it to throw an error instead. Reason is if user intented to use auth, they should import `firebase/auth` instead of just warn them.

Also I found that the test code has  some misused of chai expect throw function which should make previous auth test case for error invalided.

### Check List

- [x] All tests passing

### Relevant Issues
https://github.com/prescottprue/react-redux-firebase/issues/796

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -454,9 +454,7 @@ export const handleRedirectResult = (dispatch, firebase, authData) => {
 export const init = (dispatch, firebase) => {
   // exit if auth does not exist
   if (!firebase.auth) {
-    throw new Error(
-      '"firebase/auth" must be imported to enable authentication'
-    )
+    throw new Error('"firebase/auth" must be imported to enable authentication')
   }
   dispatch({ type: actionTypes.AUTHENTICATION_INIT_STARTED })
   // Set Auth State listener
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -454,7 +454,9 @@ export const handleRedirectResult = (dispatch, firebase, authData) => {
 export const init = (dispatch, firebase) => {
   // exit if auth does not exist
   if (!firebase.auth) {
-    return
+    throw new Error(
+      '"firebase/auth" must be imported to enable authentication'
+    )
   }
   dispatch({ type: actionTypes.AUTHENTICATION_INIT_STARTED })
   // Set Auth State listener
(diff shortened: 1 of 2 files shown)


src/actions/auth.js
```diff
@@ -454,7 +454,7 @@ export const handleRedirectResult = (dispatch, firebase, authData) => {
 export const init = (dispatch, firebase) => {
   // exit if auth does not exist
   if (!firebase.auth) {
-    return
+    throw new Error('"firebase/auth" must be imported to enable authentication')
```
prescottprue: Will this be called and throw even if the intention is to not use auth? Maybe we should throw only if `userProfile` is also set? That way the error throws if you are attempting to use a feature which requires auth, otherwise just skips the logic

test/unit/actions/auth.spec.js
```diff
@@ -73,7 +73,11 @@ describe('Actions: Auth -', () => {
       expect(onAuthStateChangedSpy).to.have.been.calledOnce
     })
     it('Errors if Firebase instance is not passed', () => {
-      expect(init(dispatch, {})).to.Throw
+      expect(() => init(dispatch, {})).to.throw()
+    })
+    it("Error if 'auth' doesn't exist in firebase", () => {
```
prescottprue: Nice catch with these!

test/unit/actions/auth.spec.js
```diff
@@ -73,7 +73,11 @@ describe('Actions: Auth -', () => {
...
+      const { auth, ...authlessFirebase } = fakeFirebase
+      expect(() => init(dispatch, authlessFirebase)).to.throw()
```
prescottprue: We could also confirm that the right message is being thrown here

src/actions/auth.js
illuminist: I just noticed that `initializeAuth` props in ReactReduxFirebaseProvider is set to `true` by default, and that also took me time to realize how I forgot to include `initializeAuth={true}` in  props and it still initialize auth for me. Which made me thought only user specified `initializeAuth={true}` will certainly have an intention to use auth feature.

Now, we have to assume `initializeAuth` will run by default. I think it could be a good idea to exhaustively check for auth related config instead.

prescottprue:
Going to close based on the comment about using config to drive this instead. Reach out if you would like to reopen

Issue #852, opened 2020-01-29
[docs] useFirestoreConnect examples are using useFirebaseConnect instead


goleary:
referenced here #806

Opening a new issue to be closed with a PR I'll be submitting shortly.

prescottprue:
Fixed in comment within src and regenerated docs. Thanks for posting!

goleary:
for future reference, it would have been nice if you'd made the changes in my PR so that I got some amount of credit for them.

Thanks for the quick action.

PR #853, opened 2020-01-29
updating `useFirestoreConnect` examples from usage of `useFirebaseConnect`

goleary:
### Description
updating `useFirestoreConnect` examples to actually use that function instead of `useFirebaseConnect`

I tested these examples with the latest versions of the libraries.

### Check List

- [x] All tests passing

### Relevant Issues
#852

### DIFF
--- a/docs/api/useFirestoreConnect.md
+++ b/docs/api/useFirestoreConnect.md
@@ -17,8 +17,8 @@ Cloud Firestore, including it's reducer, before attempting to use.
 ### Parameters
 
 -   `queriesConfigs` **([object][5] \| [string][6] \| [Array][7] \| [Function][8])** An object, string,
-    or array of object or string for paths to sync from firestore. Can also be
-    a function that returns the object, string, or array of object or string.
+  or array of object or string for paths to sync from firestore. Can also be
+  a function that returns the object, string, or array of object or string.
 
 ### Examples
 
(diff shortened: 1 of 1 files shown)


docs/api/useFirestoreConnect.md
```diff
@@ -26,39 +26,42 @@ _Basic_
 
 ```javascript
 import React from 'react'
-import { map } from 'lodash'
 import { useSelector } from 'react-redux'
-import { useFirebaseConnect } from 'react-redux-firebase'
+import { useFirestoreConnect } from 'react-redux-firebase'
```
prescottprue: These docs are actually generated by the [comments within the source code](https://github.com/prescottprue/react-redux-firebase/blob/master/src/useFirestoreConnect.js/#L21), so we will want to update them there

goleary: I see...

It would be really nice if there was a section on contributing to the docs in the repo somewhere (unless I've missed it).  It would save people from running into this issue in the future.

I'd do it myself but I'm sure I'll miss something else important @prescottprue

prescottprue:
I used your same example code and just put it directly within the comment in src then regenerated docs. Also added a section to the [contributing docs](http://react-redux-firebase.com/docs/contributing.html) about how to add to the API docs

PR #855, opened 2020-02-03
feat(types): add firestore collectionGroup type

rscotten:
### Description

Now that https://github.com/prescottprue/redux-firestore/pull/264 has been merged, this updates the Typescript types for Firestore listeners to support `collectionGroup`.

Note that this PR changes the `collection` type to be optional, because the developer can now pass in either `collection` or `collectionGroup`.  If both are accidentally omitted, the code throws an error, so we still have protection.

### Check List

- [ ] All tests passing

### Relevant Issues
https://github.com/prescottprue/redux-firestore/pull/264


prescottprue:
Thanks for making the PR so fast! I also really like the explanation of how they can both be optional - great attention to detail

PR #866, opened 2020-02-14
fix(examples): add firebase as a dependency to complete/simple example

dzoba:
### Description
If you attempt to run the simple example, you will get this error:

image

It's pretty easily solved by the change in this PR.

### Check List

- [x] All tests passing

### Relevant Issues
Not applicable.  I didn't file an issue for this.  Let me know if there needs to be one.

### DIFF
--- a/examples/complete/simple/package.json
+++ b/examples/complete/simple/package.json
@@ -9,6 +9,7 @@
     "eject": "react-scripts eject"
   },
   "dependencies": {
+    "firebase": "^7.8.2",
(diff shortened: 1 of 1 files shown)


dzoba:
Looks like there are issues beyond this, since Todos won't come in and can't be added.

image

prescottprue:
After adding the Firebase sdk I am seeing the todos load and can add one. I'll add support for the database emulator so we can see if it maybe has to do with the project itself

PR #871, opened 2020-02-26
fix(populate): handle firestore list type in populate

illuminist:
### Description
I understand that populate was designed for RTDB where it doesn't actually have list type for an array, thus when using with firestore list type(array), it outputs an object with index as a key instead of an actual array. So this little change fixes `populate` to be able to handle firestore list and it will output an array instead of plain object.

### Check List

- [x] All tests passing


prescottprue:
Amazing find! You are right that it was originally for RTDB, but glad to have it also correctly handle Firestore

PR #875, opened 2020-02-29
fix(types): switch order of schema type arguments - @macinjoke

macinjoke:
### Description

fixed this.

スクリーンショット 2020-03-01 午前0 13 46

https://github.com/prescottprue/react-redux-firebase/pull/826/files#r386035738

### Check List

- [x] All tests passing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -809,12 +809,12 @@ export function firebaseConnect<ProfileType, TInner = {}>(
 export function firebaseReducer<
   UserType,
   Schema extends Record<string, Record<string | number, string | number>>
->(state: any, action: any): FirebaseReducer.Reducer<UserType, Schema>
+>(state: any, action: any): FirebaseReducer.Reducer<Schema, UserType>
 
 export function makeFirebaseReducer<
   Schema extends Record<string, Record<string | number, string | number>>,
   UserType = {}
->(): (state: any, action: any) => FirebaseReducer.Reducer<UserType, Schema>
+>(): (state: any, action: any) => FirebaseReducer.Reducer<Schema, UserType>
 
 /**
  * React HOC that attaches/detaches Cloud Firestore listeners on mount/unmount
--- a/index.d.ts
+++ b/index.d.ts
@@ -1128,8 +1128,8 @@ export interface Data<T extends FirestoreTypes.DocumentData> {
 
 export namespace FirebaseReducer {
   export interface Reducer<
-    ProfileType = {},
-    Schema extends Record<string, Record<string | number, string | number>>
+    Schema extends Record<string, Record<string | number, string | number>>,
+    ProfileType = {}
   > {
     auth: AuthState
     profile: Profile<ProfileType>


macinjoke:
https://github.com/prescottprue/react-redux-firebase/pull/826/files#r386035738

prescottprue:
Switching these arguments was in attempt to get the [schema feature added in this PR](https://github.com/prescottprue/react-redux-firebase/pull/826) to work as expected. Is there a reason for going back? I understand that it became breaking for this switch, but seems like the right move since it was the original intention. What are your thoughts?

macinjoke:
Sorry, it was an inadequate change.
I should have also switched the parts that use `FirebaseReducer.Reducer`. So I changed `firebaseReducer` and` makeFirebaseReducer` as well. 5929bcb

The reason for ordering the `FirebaseReducer.Reducer` type arguments in the order of `Scheme`, `ProfileType` is simply that TypeScript throws an error, as shown in the screenshot image. TypeScript generics cannot bring ordinary arguments after arguments with default values. (Because the first default argument doesn't make sense)

prescottprue:
It seems like the [the ProfileType is set to an object already](https://github.com/prescottprue/react-redux-firebase/blob/master/index.d.ts/#L1104), so we should be able to remove the default setting there right?

macinjoke:
> It seems like the the ProfileType is set to an object already

It seems that the ProfileType set here has nothing to do with the ProfileType of FirebaseReducer.Reducer.
The type argument ProfileType used in generics can be either Foo or Bar.

I think the type argument ProfileType is necessary for the user to specify a type like `firebaseReducer<Foo, Bar>`. (In this case, Bar becomes the ProfileType.)

prescottprue:
Thanks for the PR. Released in [v3.2.0](https://github.com/prescottprue/react-redux-firebase/releases/tag/v3.2.0)

rscotten:
@prescottprue, @macinjoke 

Guys!  Stop changing the order!  This is going to break our code again!

@prescottprue I wish you would have asked me about this.

Just give both types defaults like this: 

```typescript
export function firebaseReducer<
  UserType,
  Schema extends Record<string, Record<string | number, string | number>>
>(state: any, action: any): FirebaseReducer.Reducer<UserType, Schema>

export function makeFirebaseReducer<
  UserType = {},
  Schema extends Record<string, Record<string | number, string | number>> = {}
>(): (state: any, action: any) => FirebaseReducer.Reducer<UserType, Schema>
```

```typescript
export namespace FirebaseReducer {
  export interface Reducer<
    ProfileType = {},
    Schema extends Record<string, Record<string | number, string | number>> = {}
  > {
```

This way nobody needs to add the `UserType` or `Schema` if they don't want to.

And `UserType` should always come first because it's more important than `Schema`.

Please make the changes I gave above or let me know and I'll submit a PR to make the change.

rscotten:
@prescottprue @macinjoke  I submitted PR #888 with the preferred approach to solving this type error.

rscotten:
@macinjoke You could have solved this problem without creating a PR with breaking changes simply by doing this:

```typescript
  FirebaseReducer.Reducer<{}, {}>
```

prescottprue:
Thanks for the PR @rscotten! For some reason I thought that is what the last PR is doing, which is why I included it in 3.2.0.

rscotten:
@prescottprue Thank you! :)

macinjoke:
> This way nobody needs to add the UserType or Schema if they don't want to.

> And UserType should always come first because it's more important than Schema.

> Please make the changes I gave above or let me know and I'll submit a PR to make the change.

> @prescottprue @macinjoke I submitted PR #888 with the preferred approach to solving this type error.



> @macinjoke You could have solved this problem without creating a PR with breaking changes simply by doing this:
> 
> ```ts
>   FirebaseReducer.Reducer<{}, {}>
> ```

Sounds great! Thank you.

JonnyBoy333:
Would someone here be kind enough to share an example of how to use the Schema part of `FirebaseReducer.Reducer(ProfileType, Schema)`? Seems like no matter what I put in there I get the error: 
`Type 'DBSchema' does not satisfy the constraint 'Record<string, Record<string | number, string | number>>'.
  Index signature is missing in type 'DBSchema'.`

Trying to do something simple like:
``` typescript
interface IContact {
  name: string;
  email: string;
}

// create schema for the DB
interface DBSchema {
  contacts: IContact
}

export interface RootState {
  firebase: FirebaseReducer.Reducer<IProfile, DBSchema>; // <--- Error here
  firestore: FirestoreReducer.Reducer;
  localSettings: ILocalSettings
  loading: ILoading
}

const appReducer = combineReducers<RootState>({
  localSettings: settingsReducer,
  loading: loadingReducer,
  firebase: firebaseReducer,
  firestore: firestoreReducer,
});
```

rscotten:
@JonnyBoy333 The original PR was #826 which broke the codebase due to putting the new `Schema` type before the `Profile` type in `FirebaseReducer.Reducer`, which was later fixed.   So make sure you're on the latest version (3.3.0).  

The schema type only works for the RealTime DB (not Firestore). I don't use the RTDB, so I can't really help you debug this.  @vicrac or @prescottprue can you help?

rscotten:
@JonnyBoy333 you can hack it by doing this:

```typescript
combineReducers<RootState>({
    firebase: firebaseReducer as any, // <-- add "as any"
  });
```

rscotten:
@JonnyBoy333 

I figured it out.  We just need to change the type definitions to:

```typescript
export function firebaseReducer<
  UserType extends Record<string, any> = {},
  Schema extends Record<string, any> = {}
>(state: any, action: any): FirebaseReducer.Reducer<UserType, Schema>

export function makeFirebaseReducer<
  UserType extends Record<string, any> = {},
  Schema extends Record<string, any> = {}
>(): (state: any, action: any) => FirebaseReducer.Reducer<UserType, Schema>
```

I'll submit a PR.

rscotten:
@JonnyBoy333 figured it out.  The type definitions need to be changed in `index.d.ts`:

```typescript
export function firebaseReducer<
  UserType extends Record<string, any> = {},
  Schema extends Record<string, any> = {}
>(state: any, action: any): FirebaseReducer.Reducer<UserType, Schema>

export function makeFirebaseReducer<
  UserType extends Record<string, any> = {},
  Schema extends Record<string, any> = {}
>(): (state: any, action: any) => FirebaseReducer.Reducer<UserType, Schema>
```

rscotten:
@JonnyBoy333 @prescottprue submitted PR #pull to fix this.

JonnyBoy333:
@rscotten Wow that was quick, thank you and also thanks for updating the documentation.

I tried copying your updated index.d.ts into my project and I am still getting an error (even with your documentation example). It's the same error as before and I think it's because there is another reducer type on line 1134. Here:
https://github.com/prescottprue/react-redux-firebase/blob/cfc516c428883a7845d6f9fc8ee671ce80e808f7/index.d.ts#L1134

rscotten:
@JonnyBoy333 Good catch.  Fixed.  Thanks!

JonnyBoy333:
@rscotten Confirmed, no more errors here.

I hope someday someone will do the same for Firestore schemas 🤞.

PR #897, opened 2020-04-01
fix(docs): correct state selector in firestore example - @gregfenton

gregfenton:
Suggested tweak to example code.  The query already limits to fetching one specific document (doc: todoId), so the result set will have just a single entry in the todos array.  I'm not even sure if ordered.todos[todoId] would even work - my brief testing shows that it would yield undefined ?

### Check List

- [ ] All tests passing

### DIFF
--- a/docs/firestore.md
+++ b/docs/firestore.md
@@ -120,7 +120,7 @@ Firestore queries can be created in two ways:
     useFirestoreConnect(() => [
       { collection: 'todos', doc: todoId } // or `todos/${props.todoId}`
     ])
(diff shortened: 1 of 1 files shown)


prescottprue:
yes it would be undefined - it should be pulling from `data` instead of `ordered` since it is looking up a value by id

gregfenton:
Or it could be ordered[0] since it has already requested a single result
with the ID query argument.

Greg Fenton.

prescottprue:
I think it may be good to use `data` since we are looking by key - just because there could be other data that has been loaded into that same path of redux by other queries

PR #906, opened 2020-04-06
fix(types): prevent type error in combineReducers by adding generic type for Schema

rscotten:
### Description

This fixes the `Schema` generic type definition for the `firebaseReducer` which caused a type error in the redux `combineReducers` function.  

It also provides an example of why and how to use the `Schema` type in the [Getting Started](https://react-redux-firebase.com/docs/getting_started.html) doc.

Minor punctuation and whitespace changes courtesy of Prettier.

### Check List

- [x] All tests passing

### Relevant Issues
#875 
#826


prescottprue:
@rscotten Would you mind taking a look to confirm everything with this works good on your end?

rscotten:
@prescottprue The `ProfileType` part definitely works.

As for the `Schema` type, it appears to work for me because VSCode gives me the expected auto-complete and errors.  Unfortunately, I don't use the Realtime DB, so I can't verify it works beyond the type-check.  But @JonnyBoy333 claims it works for him, so that's probably good enough validation.

Have you tried it on your code?

prescottprue:
Didn't realize you were the one that made the PR and wanted to have your eyes on it 😆.

Haven't gotten a chance to try it on my side yet, but should get around to it today

JonnyBoy333:
I did test that the state in my components has types for `state.firebase.data.todos`, which were inherited from the `FirebaseReducer.Reducer(Profile, Schema)`.

But in all fairness I do use Firestore so the Realtime DB typings are not very useful to me.

dhuber666:
@rscotten 
@JonnyBoy333
@prescottprue 

I think this is related to this issue I opened #912 

Do you have any idea on how to fix this?

Issue #665, opened 2019-04-01
Could upload file without  default uuid  ? 


Louis-Chen:
when i upload file 
like

```
firebase.uploadFiles('images/avatar',files,'images/avatar')
```

it will produce new uuid  and my database are ugly

firebase structure like

```
images:{
     avatar:{
          GzTbSpMI7fL1m4ae9IFY:{ 
              fullPath:"/doc/image"
              downURL:"https://www.google.com/picture"
          }
     }
}
```
but i want a single file structure

lieke

````
images:{
     avatar:{
              fullPath:"/doc/image"
              downURL:"https://www.google.com/picture"
          }
     }
}
````

could I ?

prescottprue:
You can use `uploadFile` - `uploadFiles` assumes you are passing a list of files, which requires pushing to the location provided - otherwise files in the list would run over each other.

Louis-Chen:
i use `uploadFile` and upload data to firestore but dbpath also create uuid , the function seem create `collections` when it store image data to firestore  

can`uploadFile`  create dbpath to `doc` not `collections`?

jnamkung:
@Louis-Chen Which version of react-redux-firebase are you using, 2 or 3? Curious because I am having trouble with `uploadFiles(storagePath, files, dbPath)` using version 3.

PR #936, opened 2020-05-19
feat(storage): upload file metadata to existing firestore document

JonnyBoy333:
### Description
This pull requests enhances the `uploadFile` and `uploadFiles` methods when using Firestore as your database to store the file metadata. Current behavior is to specify a path to a collection, when a file is uploaded the metadata will create a new document with that data in the specified collection. 

In some cases you may already have a document in that collection that you would like to update with the metadata instead of creating a new document. In my use case I have a contacts collection with a list of contact documents and I would like to update an existing contact document with a file download Url from the metadata instead of creating a new document with that information. This PR allows for that functionality by specifying a `documentId` property in the `uploadFile` options object. If this property exist the corresponding document will get updated, otherwise a new document will be created.

I've also updated the documentation with this new feature as well as the typescript typings. While updating the docs and typings I noticed some related properties that were missing, like the metadata and metadataFactory properties on the upload options object so I took the liberty of adding typings for those and adding them to the documentation as well. I found that to be important since they impact how my feature works.

Small side note that some of the minor changes in the documentation where made from a pre-commit linting script and not actually something that I changed, hopefully that doesn't cause confusion.

Here's the example I put in the upload examples to see it in action:
### Update Firestore Document

If using Firestore for you database and you would like to update a specific document after a file has uploaded you can specify the `options.documentId` property. In this example the document with id `12345` in the `contacts` collection will have the `fileUrl` property updated with the file's download url. More details can be found in the [upload file](https://firebase.google.com/docs/storage/web/upload-files#add_file_metadata) section of the Firebase docs.

```js
const storagePath = 'contacts/sleepygary'
const dbPath = 'contacts'

firebase
  .uploadFile(storagePath, file, dbPath, {
    metadataFactory: (uploadRes, firebase, metadata, downloadURL) => {
      return { fileUrl: downloadURL }
    },
    documentId: '12345'
  })
  .then(() => {
    console.log('File uploaded successfully')
  })
```

### Check List

- [x] All tests passing

### Relevant Issues
https://github.com/prescottprue/react-redux-firebase/issues/665


src/utils/storage.js
```diff
@@ -145,12 +145,26 @@ export function writeMetadataToDb({
...
-            .firestore()
-            .collection(dbPath)
-            .add(fileData)
+          if (documentIdFromOptions) {
+            const docRef = firebase // Write metadata to Firestore
+              .firestore()
+              .collection(dbPath)
+              .doc(documentIdFromOptions)
+            return docRef.update(fileData).then(() => docRef)
+          } else {
```
prescottprue, line 165: We shouldn't need the else here since we are returning in the if

src/utils/storage.js
```diff
@@ -145,12 +145,26 @@ export function writeMetadataToDb({
...
-          return firebase // Write metadata to Firestore
-            .firestore()
-            .collection(dbPath)
-            .add(fileData)
+          if (documentIdFromOptions) {
+            const docRef = firebase // Write metadata to Firestore
+              .firestore()
+              .collection(dbPath)
+              .doc(documentIdFromOptions)
+            return docRef.update(fileData).then(() => docRef)
```
prescottprue, line 164: Nice work! Wondering if we should use set with merge instead of update here? Or at least off the option?

index.d.ts
```diff
@@ -760,24 +754,47 @@ interface ExtendedStorageInstance {
...
-          ) => string)
-    }
+    options?: UploadFileOptions
   ) => Promise<{ uploadTaskSnapshot: StorageTypes.UploadTaskSnapshot }[]>
 }
 
+/**
+ * Configuration object passed to uploadFile and uploadFiles functions
+ */
+export interface UploadFileOptions {
```
prescottprue, line 772: Great work building this out more and separating!

index.d.ts
```diff
@@ -734,21 +734,15 @@ interface ExtendedStorageInstance {
...
    * @param options - Options
    * @param options.name - Name of the file
+   * @param options.metadata - Metadata associated with the file to upload to storage
+   * @param options.documentId - Id of document to update with metadata if using Firestore
    * @see https://react-redux-firebase.com/docs/api/storage.html#uploadFile
    */
   uploadFile: (
     path: string,
-    file: File,
+    file: File | Blob,
```
prescottprue, line 743: Glad you caught this!

src/utils/storage.js
JonnyBoy333: That's an interesting thought. Using `set` with the merge property functions the same as `update` except that in addition to updating properties on an existing document it will create a new document if the `documentId` does not already exist. I'm thinking `set` with merge should just be the default. This gives the flixibility of being able to create and update documents in bulk and puts the responsibility on the developer to make sure the documentId exists if they don't want a new document created.

I could add another option property but it kind of seems like if you're smart enough to set a flag to fail if your documentId doesn't already exist you're probably smart enough to have filtered it out of your update in the first place 😉.

I'm going to make this change.

src/utils/storage.js
JonnyBoy333: Fair enough, sometimes I think it's more readable with the `else`, but you 're right that it's not necessary. I will make this change.

JonnyBoy333:
@prescottprue  Made some updates today per your suggestions. Take a look when you have a chance. Open to feedback.

src/utils/storage.js
JonnyBoy333: I just noticed today that the `updateProfile` method has a `useSet` option that is probably what you're talking about here, maybe I will add something like that for consistency's sake.

JonnyBoy333:
Well, I did update this PR with a `useSetForMetadata` property on the uploadFiles function. This functions the same way as `useSet` on the `updateProfile` function. It defaults to true so you don't have to set it, but if for some reason you want the operation to fail if the documentId is not found you can set this to false and have that functionality. I ran some tests and made sure this is functioning as intended.

I do apologize that I'm not affluent at writing tests in mocha yet. I was hoping I could expand on existing tests around this but all the current ones seem to be around the realtime db and not firestore. I hope this doesn't block this PR.

Otherwise this PR should be ready to go with all your suggestions in place.

prescottprue:
Thanks for all of the work! Going to merge and release as part of v3.5.0

index.d.ts
rscotten: @prescottprue @JonnyBoy333 

Adding `Blob` is now breaking my type check:

```typescript
  const { uploadTaskSnapshot } = await firebase.uploadFile(
          storagePath,
          acceptedFiles[0],
          undefined,
          {
            name: (file: File) => file.name.replace(/[^\w.\-_]/gi, ''),
          }
        );
```

```
Type 'File | Blob' is not assignable to type 'File'.
            Type 'Blob' is missing the following properties from type 'File': lastModified, name ts(2345)
```

This is because the new type definition is incorrect:

```typescript
export interface UploadFileOptions {
  name?: string | ((
    file: File | Blob, // should not use the "|" operator
    internalFirebase: WithFirebaseProps<ProfileType>['firebase'],
    uploadConfig: {
      path: string,
      file: Either<File, Blob>,
      dbPath?: string,
      options?: UploadFileOptions
    }
  ) => string)
  ...
}
```

```typescript
export interface UploadFileOptions {
  name?: string | ((
    file: Either<File, Blob>, // use Either<type1, type2> instead
    internalFirebase: WithFirebaseProps<ProfileType>['firebase'],
    uploadConfig: {
      path: string,
      file: Either<File, Blob>,
      dbPath?: string,
      options?: UploadFileOptions
    }
  ) => string)
  ...
}
```

JonnyBoy333: Interesting, I just tried recreating your code and I am not getting any typescript errors. The fact that you are explicitly setting the file to a File type with `file: File` should let typescript know that it is a File and not a Blob, which allows access to the name property (I do the same thing in my code except for Blob where the name property does not exist).

``` typescript
    const testFile = new File([], 'test');
    const { uploadTaskSnapshot } = await firebase.uploadFile(
      'test/path',
      testFile,
      undefined,
      {
        name: (file: File) => file.name.replace(/[^\w.\-_]/gi, ''),
      }
    );
    // No typescript errors here
```

The reason for the union type here is because firebases file storage api does support both File and Blob types (as well as Uint8Array). And what type gets passed into the name function is dependent on the type of the second property of the `uploadFile` function.

There might be a better way to tell typescript that since a File type object was put in the `uploadFile` function then it can be certain that a File type object will get passed to the `name` function. I'd have to think about that a little more.

prescottprue: Thanks for reaching out - So guessing that should have been `Either<File, Blob>`?

rscotten: @prescottprue @JonnyBoy333 Created a PR with the fix here: #950

JonnyBoy333: @rscotten That pull request is what I was thinking of but didn't know how to do myself. Thank you! Now I'm guessing you would write your upload that like this?

``` typescript
const testFile = new File([], 'test');
const { uploadTaskSnapshot } = await firebase.uploadFile<File>( // Adding the type here
  'test/path',
  testFile,
  undefined,
  {
    name: (file) => file.name.replace(/[^\w.\-_]/gi, ''),
  }
);
```

rscotten: @JonnyBoy333 yep! That's correct.

Issue #915, opened 2020-04-18
Update Docs - Populate is now supported with Firestore


jakebern:
https://react-redux-firebase.com/docs/firestore#populate

If I understand correctly, populate works with Firestore  - but is not supported with hooks yet (https://github.com/prescottprue/react-redux-firebase/blob/master/docs/api/useFirestoreConnect.md).

jakebern:
Source suggesting was supported: https://github.com/prescottprue/redux-firestore#population

prescottprue:
Populate should work the same with hooks as it does with HOCs - are you sure you are capturing from state correctly?

Can you provide a code sample to reproduce the populate not working?

GautierT:
Hi.
Probably related to https://github.com/prescottprue/react-redux-firebase/issues/922#issue-607540672

jakebern:
@prescottprue Thanks - I was just commenting that the docs seem out of date/inconsistent: ([A](https://github.com/prescottprue/react-redux-firebase/blob/master/docs/api/useFirestoreConnect.md) vs. [B](https://github.com/prescottprue/react-redux-firebase/blob/master/docs/api/useFirestoreConnect.md)
<img width="854" alt="Screen Shot 2020-04-27 at 8 42 18 AM" src="https://user-images.githubusercontent.com/5565911/80391831-20b39000-8863-11ea-9760-e6fb6add2143.png">

I also struggled to figure out how to populate a document with objects in a sub-collection #922 (helpful Gitter thread ongoing - I haven't followed up yet). I'll follow up after doing more digging.

PR #949, opened 2020-05-30
fix(docs): cleanup minor typo in useFirestoreConnect.md

gregfenton:
QUESTION: this page still indicates that "populate is not supported".  Is this still the case for firestore?

### Check List

- [ ] All tests passing


prescottprue:
Population is supported in Firestore, but just not the full profile population

Agreed that it should be done, I think #915 is for updating the docs

PR #950, opened 2020-05-31
fix(types): fix file blob type for uploadFile and uploadFiles

rscotten:
### Description

This PR fixes a type issue in the PR #936 in which the `uploadFile()` function for file storage uploading was allowed to take a `Blob` type.  Typing the file option to have `File | Blob` broke the type checking for the `UploadFileOptions` type definition, and my solution was to make the type generic with [Conditional Types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types).

### Check List

- [ ] All tests passing

### Relevant Issues
#936


index.d.ts
```diff
@@ -10,6 +10,8 @@ import { Dispatch } from 'redux'
  */
 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
 
+type FileOrBlob<T> = T extends File ? File : Blob
```
prescottprue, line 13: Nice!

prescottprue:
Thanks for getting this in so quickly!

PR #957, opened 2020-06-04
feat(types): add database and firestore statics

msutkowski:
### Description

- Prefers extending the base interfaces for Firestore/FirebaseExtendedInstance instead of using intersections throughout the types
- Adds `FirebaseDatabaseService` types so that you can achieve:
```ts
...
const firebase = useFirebase();
firebase.database.ServerValue.TIMESTAMP
...
```
- Correct types for FirestoreStatics so you can achieve:
```ts
const firestore = useFirestore();
firestore.FieldValue.serverTimestamp()
```
_Previously, you would've gotten 'serverTimestamp() is a static member of FieldValue'_
- Fixes `useFirestore()` types - previously `withFirestore` would've been close, but `useFirestore` was missing statics.

We've also added inference for plugins, such as performance, analytics, remoteConfig, etc. When a user `import`s those plugins, TS will automatically allow a user to access those properties. This basically works by saying, 'if we see this known key on the `firebase` instance, use the types that we know firebase merged into the namespace'

### Check List

- [X] All tests passing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -357,24 +357,11 @@ type ExtendedFirebaseInstance = BaseExtendedFirebaseInstance & OptionalPick<Fire
  * @see https://react-redux-firebase.com/docs/api/firebaseInstance.html
  */
 export function createFirebaseInstance(
-  firebase: FirebaseNamespace,
+  firebase: any,
   configs: Partial<ReduxFirestoreConfig>,
   dispatch: Dispatch
 ): ExtendedFirebaseInstance;
 
-/**
- * Create an extended firebase instance that has methods attached
- * which dispatch redux actions.
- * @param firebase - Firebase instance which to extend
- * @param configs - Configuration object
- * @param dispatch - Action dispatch function
- * @see https://react-redux-firebase.com/docs/api/firebaseInstance.html
- */
-export function createFirebaseInstance(
-  firebase: any,
-  configs: Partial<ReduxFirestoreConfig>,
-  dispatch: Dispatch
-): ExtendedFirebaseInstance
 
 export type QueryParamOption =
   | 'orderByKey'
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { FirebaseNamespace } from "@firebase/app-types";
 import * as FirestoreTypes from '@firebase/firestore-types'
 import * as DatabaseTypes from '@firebase/database-types'
 import * as StorageTypes from '@firebase/storage-types'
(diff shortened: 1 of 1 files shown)


prescottprue:
This is awesome, nice work! Do you foresee this causing any issues for folks? It doesn't appear so

Asking so we can figure out the version for release

msutkowski:
@prescottprue It should not as it just 'fixes' those few TS issues. The only potential issue I can think of is in regard to not using intersections which doesn't apply here - if there was a property collision, that'd throw a type error if they differed. But as it sits, there are no overlaps here that would cause that kind of issue I can see (and we'd also see that in the build).

We could look at introducing some type tests, but being that all this is doing is setting up basic interfaces pointing to the provided Google types, I don't know if that's worth it. If you want to go down that path, I'm open to it and can get a 2nd/3rd opinion as well :)

msutkowski:
@prescottprue Side note: I just quickly scrubbed the APIs for Firebase and compared it what's currently returned by `ExtendedFirebaseInstance`. There are a few more missing elements that I can tack on to `ExtendedFirebaseInstance` quickly:

```ts
    messaging(): firebase.messaging.Messaging;
    functions(region?: string): firebase.functions.Functions;
    performance(): firebase.performance.Performance;
    analytics(): firebase.analytics.Analytics;
    remoteConfig(): firebase.remoteConfig.RemoteConfig; 
```

If you'd like to chat about any of this, I'm on the Reactiflux discord under the same user name (often found in #redux). I'm happy to contribute more if you're looking for additional help. Thanks!

msutkowski:
@prescottprue Just updated to handle those scenarios. If you include plugins, they'll automatically be inferred on the instance. Example:

```ts
import firebase from "firebase/app";
import "firebase/auth";
import "firebase/database";
import "firebase/firestore";
import "@firebase/messaging";
import "@firebase/performance";
import "@firebase/functions";
import "@firebase/analytics";
import "@firebase/remote-config";


In some component:
const firebase = useFirebase();
console.log('$', firebase.analytics(), firebase.messaging(), '...');
```

Thanks to @themindoverall for the help!

msutkowski:
@prescottprue I think this is good to go. Let me know if there's anything else you'd like done here.

index.d.ts
```diff
@@ -139,12 +140,32 @@ interface RemoveOptions {
...
+  }
+}
+
 /**
  * Firestore instance extended with methods which dispatch
  * redux actions.
  * @see https://react-redux-firebase.com/docs/api/firebaseInstance.html
  */
-interface ExtendedFirebaseInstance extends DatabaseTypes.FirebaseDatabase {
+interface BaseExtendedFirebaseInstance
```
prescottprue, line 164: Really like that you combined this all into one it much more clearly matches the name

PR #974, opened 2020-06-29
feat(populate): added support for storeAs when populating child from redux

nagarjuna993:
### Description
Added support for storeAs when populating child object from redux

### Check List

- [x] All tests passing

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -234,7 +234,7 @@ function populateChild(state, child, p) {
   // populate child is key
   if (typeof childVal === 'string' || childVal instanceof String) {
     // attach child paramter if it exists
-    const dotRoot = p.root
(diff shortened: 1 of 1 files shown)


src/helpers.js
```diff
@@ -234,7 +234,7 @@ function populateChild(state, child, p) {
   // populate child is key
   if (typeof childVal === 'string' || childVal instanceof String) {
     // attach child paramter if it exists
-    const dotRoot = p.root
+    const dotRoot = (p.storeAs ? p.storeAs : p.root)
```
prescottprue, line 237: Nit pick: This could be `p.storeAs || p.root` right? Not a big deal though, I'll still get it merged

nagarjuna993: Yes, this could've been much simpler

PR #979, opened 2020-07-10
fix(types): add populates to QueryParamOption type

bawahakim:
### Description
This is my first public Pull Request. Please let me know if I'm doing things right!

There were some types missing for Typescript : uid for Auth, and populates for firestore queries. I also wanted to add the uid to the user profile, as I find myself importing the AuthState only for the uid. Since the doc id for the user profile has to be the auth uid, I thought it would be relevant to add it to the user profile.

I am not sure if the docs should be updated for that.

Appreciate any feedback!

### Check List

- [x] All tests passing 

### DIFF
--- a/src/actions/auth.js
+++ b/src/actions/auth.js
@@ -63,7 +63,7 @@ function getProfileFromSnap(snap) {
   }
   // Firestore
   if (snap && snap.data && snap.exists) {
-    return snap.data()
+    return { ...snap.data(), uid: snap.id }
   }
   return null
 }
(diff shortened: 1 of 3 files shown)


src/actions/auth.js
```diff
@@ -63,7 +63,7 @@ function getProfileFromSnap(snap) {
   }
   // Firestore
   if (snap && snap.data && snap.exists) {
-    return snap.data()
+    return { ...snap.data(), uid: snap.id }
```
prescottprue, line 66: This may be considered a breaking change since we are changing the return type of this function which I believe returns values to a public facing function (i.e. changing the external API).

prescottprue:
Nice work! We may want to break up the two parts you are adding here

prescottprue:
In the v3.8.1 the types for `populates` were added based on your changes! This PR was not used since it also has the UID changes which appear to be breaking (change to external API)

Because of that there are now conflicts. Feel free to open a new PR or fix the conflicts for this one and update the info

PR #1008, opened 2020-09-27
feat(auth): enable claims without userProfile

rscotten:
### Description

This PR allows allows the developer to `enableClaims` without having a `userProfile`, e.g.:

```typescript
const rrfConfig: Partial<ReactReduxFirebaseConfig> &
  Partial<ReduxFirestoreConfig> = {
  enableClaims: true,
  // userProfile: 'users',
  // useFirestoreForProfile: true,
  logListenerError: process.env.NODE_ENV !== 'production',
};
```

For instance, I was previously using Firestore for my `users` collection, and that collection was at the root level.  

I had to relocate the `users` collection to be a subcollection of `customers` creating the path `customers/{customerId}/users`.

I don't know the `customerId` at the time of bootstrapping my app, so I can't do something like:

```typescript
const rrfConfig: Partial<ReactReduxFirebaseConfig> &
  Partial<ReduxFirestoreConfig> = {
  userProfile: `customers/${customerId}/users`, // Not possible
  useFirestoreForProfile: true,
  logListenerError: process.env.NODE_ENV !== 'production',
};
```

But I can store the `customerId` in the Firebase Auth custom claims, and use that to determine the customer account the user can access.  Hence this PR.

### Check List
If not relevant to pull request, check off as complete

- [x] All tests passing
- [x] Docs updated with any changes or examples if applicable
- [x] Added tests to ensure new feature(s) work properly

PR #1041, opened 2020-12-14
fix(auth): unWatchUserProfile error when database is not setup

ssdns:
### Description
My configuration is  ```userProfile:  null,  enableClaims: true``` (ref: https://github.com/prescottprue/react-redux-firebase/pull/1008) and without setup firebase/database.
It throws "firebase.database is not function" error in firebase.logout.

### Check List

- [x] All tests passing


ssdns:
:bow:

PR #1052, opened 2021-01-05
fix(docs): correctly redirect when auth does not exist in the routing example doc

JonathanPorta:
The current example routes a user to the `/login` page if they are logged in.

Based on the context of this code, I am pretty sure the intended functionality is to show how to bounce a user to the `/login` route when `NOT` logged in.

### Check List

- [x] All tests passing

### DIFF
--- a/docs/recipes/routing.md
+++ b/docs/recipes/routing.md
@@ -18,7 +18,7 @@ class ProtectedPage extends Component {
     authExists: PropTypes.bool,
   }
 
-  componentWillReceiveProps({ authExists }) {
+  componentWillReceiveProps({ !authExists }) {
     if (authExists) {
       this.context.router.push('/login') // redirect to /login if not authed
     }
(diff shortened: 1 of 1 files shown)


docs/recipes/routing.md
```diff
@@ -18,7 +18,7 @@ class ProtectedPage extends Component {
     authExists: PropTypes.bool,
   }
 
-  componentWillReceiveProps({ authExists }) {
+  componentWillReceiveProps({ !authExists }) {
```
prescottprue, line 21: This is destructuring off of props - we would want to move this down to line 22 instead of here

docs/recipes/routing.md
```diff
@@ -18,7 +18,7 @@ class ProtectedPage extends Component {
...
     if (authExists) {
```
prescottprue, line 22:
```suggestion
    if (!authExists) {
```

docs/recipes/routing.md
prescottprue, line 21:
```suggestion
  componentWillReceiveProps({ authExists }) {
```

prescottprue:
Added to https://github.com/prescottprue/react-redux-firebase/pull/1149

Issue #1073, opened 2021-03-16
Minor: Typo at Constants docs


MorenoMdz:
At [the Auth page](https://react-redux-firebase.com/docs/auth.html), logout() section, an s is missing from the **constants** page link. 

> logout()
> Logout from Firebase and delete all data from the store (state.firebase.data and state.firebase.auth are set to null).
> 
> Looking to preserve data on logout? **checkout the preserve config option**, which preserves data under the specified keys in state on logout.

Poiting to `https://react-redux-firebase.com/docs/api/**contants**` instead.

codedpills:
@MorenoMdz Good catch. I've submitted a PR to patch this. Thanks

PR #1086, opened 2021-04-25
fix(docs): broken link caused by typo - #1073

codedpills:
### Description
At the Auth page, logout() section, the link to the constants page is broken because an "s" is missing from the constants page link.

### Check List

- [ ] All tests passing


prescottprue:
Great catch, thanks for the PR!

PR #1095, opened 2021-05-08
feat(types): add generic to FirestoreReducer Reducer interface

SamPoate:
### Description
Adds the ability to define the interface for FirestoreReducer Reducer interface.

I am new to this, so please do leave any feedback.

### Check List

- [ N/A ] All tests passing
- [ N/A ] Docs updated with any changes or examples if applicable
- [ N/A ] Added tests to ensure new feature(s) work properly


prescottprue:
Thanks for the change, I really like the idea! I thought we looked into something like this before, but can't find it.

Unless I'm misunderstanding, I believe this will require a type be passed to the reducer, which means a breaking change to the external API. I'll put this on the list for a next major version, but if we can think of a way to support this in a backwards compatible way we can get it in!

